PR #1, opened 2018-07-29
Handles memfd file descriptors

fdumontmd:
A memfd file descriptor looks like `/memfd:pulseaudio (deleted)`; the current code panics when trying
to convert the name into a `u32`.

This change adds support for such entries.


eminence:
Thanks!

Issue #2, opened 2018-08-15
Improving is_alive()


levex:
Hi Andrew,

First of all - thanks for the great library, we use it in production for a larger Rust-based system supervisor project.

That said, I have two issues with `is_alive()`:
 - Is a defunct (as told by the `Z` flag in `/proc/$pid/stat`) process "alive"? Probably not :-)
 - We have a system where we've hit issues with `pid` reuse, while the `cmdline` and `uid` are the same. `is_alive()` seems to agree that they are same process, but in reality they aren't.

What do you think about a PR that would:
 - Check whether the process is defunct before returning `true` in `is_alive()`?; and
 - Cache `starttime` from `/proc/$pid/stat` and compare the current value to determine if they are actually the same process.

Note that using `starttime` is still not bullet-proof, but for my use case it's a better check.

Thanks!

Lev

eminence:
Hi Lev,

I'm glad to hear that this library is useful in your project.  Thanks!

About your is_alive issues: your ideas sound great, and a PR would be happily accepted!   The `starttime` field will already be available, since the `Stat` structure is populated when a `Process` struct is created.  So this will be an easy check to make.

PR #3, opened 2018-08-15
process: improve is_alive by checking starttime and state

levex:
Checking /proc/$pid/stat:starttime to determine whether the process
is the same as the one we cached decreases the chance of misinterpreting
the state of the process. While at it, also make sure that processes that
are zombies ("defunct") are not shown as alive.

Closes: #2 
Signed-off-by: Levente Kurusa <lkurusa@acm.org>


eminence:
Thanks!  Merged and published as v0.3.1

PR #4, opened 2018-09-10
use kibibytes to parse meminfo file

flier:
As you known, `/proc/meminfo` use `KB` which means kilobytes, but it's typo to kibibyte as [RedHat mentioned](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/s2-proc-meminfo). 

> While the file shows kilobytes (kB; 1 kB equals 1000 B), it is actually kibibytes (KiB; 1 KiB equals 1024 B). This imprecision in /proc/meminfo is known, but is not corrected due to legacy concerns - programs rely on /proc/meminfo to specify size with the "kB" string.

There also has [a kernel patch](https://lore.kernel.org/patchwork/patch/666444/) for the issue, but it has been rejected

> Does the difference actually matters so much that we should change a
user visible format of a file? Some users might not ready to changes
(say sombody did sed 's@.*:[[:space:]]*\([0-9]*\) kB@\1@' to get values.
This change would break it which is something we try to prevent as much
as possible.
>
> So I do not think this all is worth the potential troubles. There are
probably other places where we present kB while we in fact think kiB.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -284,6 +284,16 @@ fn convert_to_bytes(num: u64, unit: &str) -> u64 {
     }
 }
 
+fn convert_to_kibibytes(num: u64, unit: &str) -> u64 {
+    match unit {
+        "B" => num,
+        "KiB" | "kiB" | "kB" | "KB" => num * 1024,
+        "MiB" | "miB" | "MB" | "mB" => num * 1024 * 1024,
+        "GiB" | "giB" | "GB" | "gB" => num * 1024 * 1024 * 1024,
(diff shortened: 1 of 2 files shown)


eminence:
Thanks!   Do you think you could add a doccomment to the MemInfo struct with this info?   Something simple like "Note:  Even though the /proc/meminfo file uses the "kB" unit, the values are actually kibibytes, not kilobytes".

flier:
sure, I have add a note about this

eminence:
Thanks!

PR #5, opened 2018-09-10
read kernel config from /boot

flier:
As you known, the kernel config may not exists at `/proc/config.gz` on some Linux release, we need to check the following location.

* /proc/config.gz
* /boot/config
* /boot/config-$(uname -r)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -545,9 +545,6 @@ pub fn kernel_config() -> ProcResult<HashMap<String, ConfigSetting>> {
         let file = proctry!(File::open(PROC_CONFIG_GZ));
         let decoder = proctry!(Decoder::new(file));
         Box::new(BufReader::new(decoder))
-    } else if Path::new(BOOT_CONFIG).exists() {
-        let file = proctry!(File::open(BOOT_CONFIG));
-        Box::new(BufReader::new(file))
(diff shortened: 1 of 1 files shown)


eminence:
Thanks for the suggestion.  I like the idea, but I'm sure I like the magic of searching for things in /boot/.  If /proc/config.gz exists, it's guaranteed to match the running kernel.  But /boot/config might not.

I would instead prefer that kernel_config() continue to read only from /proc/config.gz, but we can add another function that'll accept a file path (like "/boot/config") and use the same config parsing code.

flier:
I agree `/boot/config` may not match the running kernel, but `/boot/config-kernel-version` [should be](https://www.tldp.org/LDP/Linux-Filesystem-Hierarchy/html/boot.html), we can adjust the order to try read it first.

On the other hand, some Linux release haven't `/proc/config.gz` by default, for example, centos 7, ubuntu 18 etc.

eminence:
Ok, sure.  I can live with that compromise.  Thanks.

flier:
thanks

eminence:
I've just published version `v0.3.2` to crates.io, which contains all 3 of your fixes.

flier:
great!

PR #8, opened 2018-09-14
Added rx_queue and tx_queue to TcpNetEntry

tvannahl:
I added the `rx_queue` and `tx_queue` ready to be ready to use. The
tests are running as fine as previously as well as the netstat example.

I do hope the formating is fine.


eminence:
Thanks!

PR #10, opened 2018-09-18
Switch away from the custom ProcResult type to a standard Result type

eminence:
I had some extra free time, so I worked on this.  CC @tvannahl -- how does this look?


eminence:
Thanks for the review!  I've merged this, but haven't yet published a new version to crates.io yet

tvannahl:
Thanks a lot. I'll play around with it further on and talk to you if I miss or see something.

PR #12, opened 2018-10-03
WIP: /proc/<pid>/mountstats parsing

tvannahl:
After an initial [conversation on twitter](https://twitter.com/thatcks/status/1046831996293566464) with  Chris Siebenmann about `/proc/<pid>/moutstats` it appears to be a good time to implement a parser for that into `procfs`. 

The basics are pretty easy to parse, but if you have an NFS filesystem the topic not as good documented. Fortunately Chris Siebenmann has written some [blog posts](https://utcc.utoronto.ca/~cks/space/blog/linux/NFSMountstatsIndex) about that a while back.

I've already done some work on the data structures and written a basic test.

# Overview Todo

 1. Finish and validate the datastructures
 2. Include documentation for every field since most fields are not coverd by `proc(5)`
 3. Implement a parser for regular filesystems
 4. Implement a parser for the NFS stat formats `1.0` and `1.1`

# Questions

 1. Some fields (like `opts`) do have a key-value, key-only structure. How should those field be handled? I do see three possible option about how to handle that:
    1. Provide those with `String` only. This is easy and the user needs to manage the parsing according to his own needs
    2. `Vec<String>` in which the settings are already split by `,`. This provides in my opinion no advantage over the first option
    3. A key-value data structure where its possible to loopup if a setting is **present**, if the setting **has a value** and the **content** of the value. My first impression was to use a `HashMap<String, Option<String>>` but I am unsure about the ergonomics. Take the following example into account:

```rust
// opts containing "rw,vers=4.1,rsize=131072"
let opts: HashMap<String, Option<String>> = stat.opts;
// key only setting
match opts.get("rw") {
    Some(None) => println!("rw is set"),
    None => println!("rw is not set"),
    Some(Some(_)) => panic!("rw should have no values"),
}
// read key-value setting
match opts.get("vers") {
    Some(Some(vers)) => println!("NFS version: {}", vers),
    Some(None) => panic!("expected version in 'vers' option"),
    None => println!("NFS version is not explicitly set"),
}
```
  2. The fitting integer values for many values are currently missing. I guess there is some kernel ready in order to determine that.

  3. Regarding the `per-op statistics`: My current approach is a relative general parsing aproach, it would also possible to define a type for every `per-op` line like `NULL`. It seems that every `statvers` does provide a different set of option. I would prefere the general aproach, but maybe there are other opinions.

### DIFF
--- a/src/process.rs
+++ b/src/process.rs
@@ -448,7 +448,8 @@ pub struct Io {
 }
 
 /// Mount information from `/proc/<pid>/mountstats`.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, Clone)]
+#[cfg_attr(test, derive(PartialEq))]
 pub struct MountStat {
     pub device: Option<String>,
     pub mount_point: PathBuf,
@@ -498,7 +499,8 @@ impl MountStat {
 }
 
 /// Only NFS mounts provide additional statistics in `MountStat` entries.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, Clone)]
+#[cfg_attr(test, derive(PartialEq))]
 pub struct MountNFSStatistics {
     pub version: String,
     // * opts: HashMap<String, Some(String)>
@@ -570,7 +572,8 @@ impl MountNFSStatistics {
 /// The underlying data structure in the kernel can be found under *fs/nfs/iostat.h* `nfs_iostat`.
 /// The fields are documented in the kernel source only under *include/linux/nfs_iostat.h* `enum
 /// nfs_stat_eventcounters`.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, Copy, Clone)]
+#[cfg_attr(test, derive(PartialEq))]
 pub struct NFSEventCounter {
     inode_revalidate: libc::c_ulong,
     deny_try_revalidate: libc::c_ulong,
@@ -642,7 +645,8 @@ impl NFSEventCounter {
 /// The underlying data structure in the kernel can be found under *fs/nfs/iostat.h* `nfs_iostat`.
 /// The fields are documented in the kernel source only under *include/linux/nfs_iostat.h* `enum
 /// nfs_stat_bytecounters`
-#[derive(Debug, PartialEq)]
+#[derive(Debug, Copy, Clone)]
+#[cfg_attr(test, derive(PartialEq))]
 pub struct NFSByteCounter {
     pub normal_read: libc::c_ulonglong,
     pub normal_write: libc::c_ulonglong,
@@ -700,7 +704,8 @@ impl NFSByteCounter {
 /// >  are measured.
 ///
 /// (source: *include/linux/sunrpc/metrics.h* `struct rpc_iostats`)
-#[derive(Debug, PartialEq)]
+#[derive(Debug, Clone)]
+#[cfg_attr(test, derive(PartialEq))]
 pub struct NFSOperationStat {
     /// Count of rpc operations.
     pub operations: libc::c_ulong,
--- a/src/process.rs
+++ b/src/process.rs
@@ -450,10 +450,10 @@ pub struct Io {
 /// Mount information from `/proc/<pid>/mountstats`.
 #[derive(Debug, PartialEq)]
 pub struct MountStat {
-    device: Option<String>,
-    mount_point: PathBuf,
-    fs: String,
-    statistics: Option<MountNFSStatistics>,
+    pub device: Option<String>,
+    pub mount_point: PathBuf,
+    pub fs: String,
+    pub statistics: Option<MountNFSStatistics>,
 }
 
 impl MountStat {
@@ -500,10 +500,10 @@ impl MountStat {
 /// Only NFS mounts provide additional statistics in `MountStat` entries.
 #[derive(Debug, PartialEq)]
 pub struct MountNFSStatistics {
-    version: String,
+    pub version: String,
     // * opts: HashMap<String, Some(String)>
     /// Duration the NFS mount has been in existence.
-    age: Duration,
+    pub age: Duration,
     // * fsc (?)
     // * impl_id (NFSv4): Option<HashMap<String, Some(String)>>
     // * caps: HashMap<String, String>
@@ -514,7 +514,7 @@ pub struct MountNFSStatistics {
     // * RPC iostats version:
     // * xprt
     // * per-op statistics
-    per_op_stats: NFSPerOpStats,
+    pub per_op_stats: NFSPerOpStats,
 }
 
 impl MountNFSStatistics {
--- a/src/process.rs
+++ b/src/process.rs
@@ -8,6 +8,7 @@ use std::io::{self, Read};
 use std::os::linux::fs::MetadataExt;
 use std::path::PathBuf;
 use std::str::FromStr;
+use std::time::Duration;
 
 // provide a type-compatible st_uid for windows
 #[cfg(windows)]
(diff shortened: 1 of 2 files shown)


eminence:
Thanks for starting this discussion!  I'm excited to add support for this.  Some additional thoughts on your question:

> Some fields (like opts) do have a key-value, key-only structure. How should those field be handled?

I agree with your thinking here.  `HashMap<Option<String>>` seems nice, but the double-unwrap isn't the best.  I'm not sure what's best, so for now, let's start by just storing opts as a `String` or `Vec<String>`.   This crate still hasn't reached the 1.0.0 semver milestone, so I'm OK with evolving this API later.

> The fitting integer values for many values are currently missing. 

Sorry, I didn't understand this.  What are "fitter interger values"?

> It seems that every statvers does provide a different set of option. I would prefere the general aproach, but maybe there are other opinions.

Agreed.

src/process.rs
```diff
@@ -446,6 +447,99 @@ pub struct Io {
...
+pub struct MountNFSStatistics {
+    version: String,
+    // * opts: HashMap<String, Some(String)>
+    /// Duration the NFS mount has been in existence.
+    age: Duration,
+    // * fsc (?)
+    // * impl_id (NFSv4): Option<HashMap<String, Some(String)>>
+    // * caps: HashMap<String, String>
+    // * nfsv4 (NFSv4): Option<HashMap<String, Some(String)>>
+    // * sec: HashMap<String, Some(String)>
```
eminence: Probably `opts`, `impl_id`, `caps`, `nfsv4`, and `sec` should all be the same type

src/process.rs
```diff
@@ -446,6 +447,99 @@ pub struct Io {
...
+    events: NFSEventCounter,
+    bytes: NFSByteCounter,
+    // * RPC iostats version:
+    // * xprt
+    // * per-op statistics
+    per_op_stats: NFSPerOpStats,
+}
+
+#[derive(Debug, PartialEq)]
```
eminence: Why `PartialEq`?

src/process.rs
```diff
@@ -446,6 +447,99 @@ pub struct Io {
...
+    // * operations
+    // * transmissions
+    // * major timeouts
+    // * bytes sent
+    // * bytes received
+    // * cumulative response time
+    // * cumulative total request time
+}
+
+pub type NFSPerOpStats = HashMap<String, NFSOperationStat>;
```
eminence: This looks good

src/process.rs
tvannahl: Yeah I was thinking about something like `type Options = HashMap<String, Option<String>>` but I am not really fond of the name yet.

src/process.rs
tvannahl: Basically two reasons:

 * `PartialEq`is required for the tests to compare the result
 * `Eq` can be added since I see no reason why this is not reflexive. But `Eq` doesn't add any functions, so `PartialEq` is still required.

eminence: Ah, good point about needed to compare them in tests.   Does it make sense to derive `PartialEq` only when we're running tests?   My only concern (which is pretty minor) is that a public PartialEq derivation suggests that it makes sense for a user to compare these objects, which I don't think is true.

While on the same subject, I wonder if these structs should get `Copy` and `Clone`

tvannahl:
> Thanks for starting this discussion! I'm excited to add support for this. 

So am I :smile: 

> 
> > The fitting integer values for many values are currently missing.
> 
> Sorry, I didn't understand this. What are "fitter interger values"?
> 

Sorry, I guess I picked the wrong vocabulary here. I meant the matching integer values, so those that are used by the kernel to generate those entries. I see that you have chosen `u64`, I assume that is safe to use and probably for fun to handle later anyways.

src/process.rs
tvannahl: That is a good point, I'll add it to the Todo-list

tvannahl:
> > Some fields (like opts) do have a key-value, key-only structure. How should those field be handled?
> 
> I agree with your thinking here. `HashMap<Option<String>>` seems nice, but the double-unwrap isn't the best. I'm not sure what's best, so for now, let's start by just storing opts as a `String` or `Vec<String>`. This crate still hasn't reached the 1.0.0 semver milestone, so I'm OK with evolving this API later.

That is a good point. I even missed one case on which the values can be a number. To parse that correctly it would require to have a deeper knowledge about every possible option within the kernel. Due to that I would now tend to `String` or `Vec<String>` as well since the end program needs to provide a deeper knowledge anyway if this is used. But it would at least great if this is consistent with other parts that do provide a simillar line like `/proc/cmdline` or `/proc/[pid]/cmdline`. So either a raw value or a `Vec<String>`.

eminence:
Ok.  Let's go with `Vec<String>`.

src/process.rs
```diff
@@ -446,6 +447,248 @@ pub struct Io {
...
+                };
+
+                v.push(MountStat {
+                    device,
+                    mount_point,
+                    fs,
+                    statistics
+                });
+            } else {
+                println!("Skipping line {}", line);
```
eminence: Reminder to myself to remove this debugging statement I added my accident.

src/process.rs
```diff
@@ -448,7 +448,8 @@ pub struct Io {
 }
 
 /// Mount information from `/proc/<pid>/mountstats`.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, Clone)]
+#[cfg_attr(test, derive(PartialEq))]
```
tvannahl: That one I like :smiley: … I didn't know about that yet.

eminence:
I'm going some more work on parsing and testing, and I've found this interesting case on my server:

one of the per-op stats looks like this:

   GETATTR: 9045698 9045704 2 2134781060 2460416188 18446744062969033049 2702499 3654112

The 6th field is the "cumulative queue time" field.  first observation is that it's really big.  It overflows the current parsing because we try to parse it as i64.  I've changed this to u64 for now.

Chris blog posts document this field as "How long (in milliseconds) all requests spent queued for transmission before they were sent".  But even if i divide this very large number by the number of transmissions (the 2nd field), the result still seems too large (average 64 years queue time per transmission!).

So something's not right here.

eminence:
The `mountstats` package from nfs-utils reports similar data, so maybe there's a kernel bug?

```
GETATTR:
        9807220 ops (3%)        6 retrans (0%)  2 major timeouts
        avg bytes sent per op: 235      avg bytes received per op: 271
        backlog wait: 1880935072627.010986      RTT: 0.290362   total execute time: 0.394107 (milliseconds)
CLOSE:
        4277803 ops (1%)
        avg bytes sent per op: 259      avg bytes received per op: 175
        backlog wait: 0.031240  RTT: 0.338436   total execute time: 0.385167 (milliseconds)
LOOKUP:
        4128123 ops (1%)
        avg bytes sent per op: 273      avg bytes received per op: 232
        backlog wait: 4468554849801.537109      RTT: 0.241026   total execute time: 0.271000 (milliseconds)
```

tvannahl:
This really looks like a ridiculously high value. I've asked Chris Siebenmann to take a look - as he probably knows the fields a little longer. The only thing that comes into my mind is the assumption that the system clock has changed or if you have a lot of messages that never where transmitted.

But there seems to be an error on my data type analysis as well if those numbers can appear in the output. What I've done to figure out the size of the basic data type is the following:

 1. The datatype of the fields `om_queue`, `om_rtt` and `om_execute` in `rpc_iostat` is `ktime_t`
 2. The function to print those values can be found in `net/sunrpc/stats.c` namely ` _print_rpc_iostats`
 3. The used macro `ktime_to_ms` does return the `s64` type (defined in `include/linux/ktime.h`)
 4. `s64` is defined in `include/acpi/actypes.h` like the following:

 > 1) The following types are of fixed size for all targets (16/32/64):
 > u8           Logical boolean
 > 
 > u8           8-bit  (1 byte) unsigned value
 > u16          16-bit (2 byte) unsigned value
 > u32          32-bit (4 byte) unsigned value
 > u64          64-bit (8 byte) unsigned value
 >
 > s16          16-bit (2 byte) signed value
 > s32          32-bit (4 byte) signed value
 > s64          64-bit (8 byte) signed value

So even though the format string in ` _print_rpc_iostats` uses `%Lu` the possible uppor bound should be limited to an `s64`.

tvannahl:
I got a message from Chris Siebenmann [on twitter](https://twitter.com/thatcks/status/1049447136797057024):

> The short answer about the stats is that something has clearly gone peculiar in the kernel. As far as I  can tell, the cumulative queue time should always be less than the cumulative total RPC time, and it’s  clearly not in your example.

This got me thinking: I got it all wrong. The essential error was when I gave the type a greater significance than the print format. Even so the format is signed, the print is not. So using `u64` to parse the print of a `ktime_t` here is the right choice for us on all those parameters. Internally in the kernel you've got yourself a negative `ktime_t` `-10740518567` after `ktime_to_ms` to be exact.

eminence:
Ok, after a bit of a break -- time to come back to this work :)  I'm going to go ahead and merge this.  I know we've not parsing every last field, but I think this is a great start, and if people have a need for something that we're not parsing right now, they can let us know via a new issue.

Thanks again for your help with this!

PR #19, opened 2019-01-29
Add udp/udp6

dalance:
I added interfaces for udp/udp6.
`UdpNetEntry` is the same as `TcpNetEntry` without `state` which seems to be meaningless.
Function `udp` and `udp6` correspond to `tcp` and `tcp6`.

### DIFF
--- a/src/net.rs
+++ b/src/net.rs
@@ -52,6 +52,16 @@ pub struct TcpNetEntry {
     pub inode: u32,
 }
 
+/// An entry in the UDP socket table
+#[derive(Debug)]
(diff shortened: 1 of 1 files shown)


eminence:
Thanks!

It does seem weird to me that there is a non-empty state field, but I wasn't able to find any details on it with a quick skim.   I'd love to find details on it, but ignoring it (as you've done here) is the right choice for right now.

eminence:
This is now published on crates.io as v0.4.3

dalance:
Thanks!

PR #20, opened 2019-02-07
Fix <pid>, <tid> dissapeared in rustdoc

dalance:
In rustdoc, some `<pid>` and `<tid>` is not rendered like below:

```
signal: u64
[−]
The bitmap of pending signals, displayed as a decimal number. Obsolete, because it does not provide information on real-time signals; use /proc//status instead.
```

https://docs.rs/procfs/0.4.3/procfs/struct.Stat.html#structfield.signal


eminence:
Thanks!

PR #22, opened 2019-02-08
Support `/proc/[pid]/status`

dalance:
I added `struct Status` and `Process::status()` corresponded to `/proc/[pid]/status`.

I referred to `Io` implementation. But I commented out `map.is_empty()` check.
This is because some patches added at new kernel are applied to older kernel modified by distribution.
For example, `umask` is added since kernel 4.7, but kernel-3.10.0-693.el7 by Red Hat has `umask`.
I followed original kernel version because I can't know what patch is applied to kernels of each distribution.


eminence:
Looks great, thanks!

Since the info in `/proc/pid/status` contains some duplicate data from `/proc/pid/stat`, what do you think about some kind of test that compares the two and makes sure they are identical?    One potential problem I see with this idea is that it's impossible to load both files at exactly the same time, and so it might be hard to compare them (their data might be slightly differernt)

eminence:
> But I commented out `map.is_empty()` check.  This is because some patches added at new kernel are applied to older kernel modified by distribution.

Sounds good to me 👍

dalance:
> Since the info in `/proc/pid/status` contains some duplicate data from `/proc/pid/stat`, what do you think about some kind of test that compares the two and makes sure they are identical? One potential problem I see with this idea is that it's impossible to load both files at exactly the same time, and so it might be hard to compare them (their data might be slightly differernt)

I added some assertions. I tried to check dynamic entry like `vmrss`, but there is difference.
So some static entries are added.

eminence:
Thanks!  Sorry for the delay

dalance:
Thanks!

eminence:
This is now published as `v0.4.4`

PR #23, opened 2019-02-12
Fix Status::vm* missing for kernel thread

dalance:
Kernel thread like `kthreadd`, `ksoftirqd`,,, don't have `Vm*` entries in `/proc/*/status`.
So all these entries changed to `Option` and added test for `kthreadd`.


eminence:
Thanks!

dalance:
Could you publish this version?

eminence:
Done!

dalance:
Thanks!

PR #24, opened 2019-02-16
Remove kernel version check from Status

dalance:
Some optional entries are not shown with supported kernel.
For example, `Umask` is not shown in zombie process's status.
If kernel config `CONFIG_PID_NS` is disabled, `NS*` is not shown.
So I think kernel version check of optional entry can be removed
and the existence should be judged by status string only.


eminence:
Sounds good to me

dalance:
Thanks!

PR #26, opened 2019-03-01
Change inode to u64

dalance:
Process::maps() call becomes panic in XFS.
This is because XFS partition over 1TB has 64bit inode.
So inode should be u64.


eminence:
Thanks!  Published as `v0.4.7`

dalance:
Thanks!

PR #28, opened 2019-03-18
Improve `dump` example

ErichDonGubler:
See commit list for more details. Feel free to reject the commits that aren't interesting -- these are definitely subjective improvements. :)


eminence:
Looks ok to me!  Thanks

PR #30, opened 2019-03-28
Use `c_char`, which is the proper alias to use with `CStr`

ErichDonGubler:
See [`std` docs here](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_ptr).

Just using `i8` breaks things like `aarch64`, which actually have a different definition of `char` in C. :) And yes, that means I'm actually participating in a project where this is getting ported to embedded devices!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -140,6 +140,7 @@ use std::fmt;
 use std::fs::File;
 use std::io::{Read, Write};
 use std::mem;
+use std::os::raw::c_char;
 use std::path::Path;
 use std::str::FromStr;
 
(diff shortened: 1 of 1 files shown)


ErichDonGubler:
EDIT: Updated OP with some more info. And hopefully something cool for you as a maintainer! :)

eminence:
Thanks!  BTW, are you running on a real aarch64 machine?  or some emulated that?  I'm asking because I'd be interested in learning how to test on other platforms than amd64 (and testing with travis CI, if at all possible).

ErichDonGubler:
We're on a real one! I'd definitely suggest you take a look at [`trust`](https://github.com/japaric/trust/) for testing on different architectures and platforms. It's a Docker-heavy flow, but it's a great solution!

Issue #29, opened 2019-03-23
Improve the ProcError


vorner:
Hello

I've noticed the ProcError doesn't implement `std::error::Error`, nor does it implement `Display`. This is annoying, because it can't be directly used with common error handling strategies (eg. can't be used as `Box<dyn Error + Send + Sync>` or with `failure`).

Furthermore, the `NotFound` variant is not very explaining about *what* was not found, which makes tracking the origin of the error or the cause a bit harder :-(.

Is it possible to improve the error type?

Thank you

eminence:
Good suggestions.

Regarding the `NotFound` variant, this can be raised if we think that a certain file should exist (basically anything under `/proc`, like `/proc/self/maps`), but this file doesn't exist.   If you know what function returned this error, generally the docs for that function should say that file it's trying to read.

What if the `NotFound` variant included the a PathBuf to the thing it was trying to open?  Would that be helpful? 

Also, Implementing `Error` and `Display` is a good idea, I'll do that (or a PR would be welcome if you wanted to do it)

vorner:
>  If you know what function returned this error, generally the docs for that function should say that file it's trying to read.

I as the programmer might track that down, using the source code and the docs. But an administrator reading through logs where it ended up as a warning or some other user might have hard time.

> What if the NotFound variant included the a PathBuf to the thing it was trying to open? Would that be helpful?

Yes, definitely.

PR #32, opened 2019-04-03
ProcError improvements

eminence:
This implements the suggestions made by @vorner in #29:

* `ProcError` variants now contain the path that caused the error
* `ProcError` implements `std::error::Error` (so it's now usable with the failure crate).  This also means it implements `Display` now.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -344,15 +344,14 @@ fn split_into_num<T: FromStrRadix>(s: &str, sep: char, radix: u32) -> (T, T) {
     (a, b)
 }
 
-
 /// This is used to hold both an IO error as well as the path of the file that originated the error
 #[derive(Debug)]
 struct IoErrorWrapper {
     path: PathBuf,
-    inner: Option<Box<dyn std::error::Error + Send + Sync>>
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -344,38 +344,151 @@ fn split_into_num<T: FromStrRadix>(s: &str, sep: char, radix: u32) -> (T, T) {
...
+    fn read_to_end(&mut self, buf: &mut Vec<u8>) -> io::Result<usize> {
+        wrap_io_error!(self.path, self.inner.read_to_end(buf))
+    }
+    fn read_to_string(&mut self, buf: &mut String) -> io::Result<usize> {
+        wrap_io_error!(self.path, self.inner.read_to_string(buf))
+    }
+    fn read_exact(&mut self, buf: &mut [u8]) -> io::Result<()> {
+        wrap_io_error!(self.path, self.inner.read_exact(buf))
+    }
+}
```
vorner: Cool idea.

vorner:
I've only skimmed it, but it looks nice. Thank you.

PR #45, opened 2019-10-07
Fix kernel version parsing

dalance:
Failure to parse kernel version (5.2.19_1) is reported at https://github.com/dalance/procs/issues/25.

### DIFF
--- a/src/sys/kernel.rs
+++ b/src/sys/kernel.rs
@@ -45,8 +45,13 @@ impl Version {
     ///
     /// ```
     pub fn from_str(s: &str) -> Result<Self, &'static str> {
-        let mut s = s.split('-');
-        let kernel = s.next().unwrap();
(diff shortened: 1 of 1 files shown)


src/sys/kernel.rs
```diff
@@ -45,8 +45,13 @@ impl Version {
     ///
     /// ```
     pub fn from_str(s: &str) -> Result<Self, &'static str> {
-        let mut s = s.split('-');
-        let kernel = s.next().unwrap();
+        let pos = s.find(|c: char| c != '.' && ((c < '0') || (c > '9')));
```
eminence: Could `! c.is_ascii_digit()` be used here?  It might be slightly more readable.

dalance: Thanks. That's right.

eminence:
Thanks for the quick fix.  This has been published as v0.5.4

PR #47, opened 2019-10-11
Add loginuid parsing to process

idanski:


eminence:
LGTM.  Only 2 comments:

* The comment says "since Linux 4.18.0", but this file seems to be present (or was backported) into much older kernels (including some 2.6.x kernels).  
* The new test will fail if it's run on a kernel that doesn't have this file.  But this isn't really a big deal since the test cases are mainly for developers, and loginuid is present on all of my development systems.  If any developers find themselves working on a system without this file, the test will have to be changed

idanski:
Thanks @eminence .
* I'll fix the test in another PR later today.
* I'm not sure how to change the comment, I understand this feature was first introduced on Linux 4.18 but I don't know to which versions it was backported to, how should I change the comment?

PR #48, opened 2019-10-14
Remove as many panics as possible

### LINKED ISSUE #44: Turn procfs into a panic-free library
In the current design of the procfs crate, panics were possible, as assertions and unwraps are common in the code.  This was done to help uncover bugs in the library.

As the library has matured a bit, all of the assertions and unwraps should be replaced with `Result` types.  Most functions already return a `ProcResult`, so this itself isn't a big change to the public API of procfs.  

See also #39

eminence:
The `ProcError` error type now gains a new `InternalError` variant which
is designed to cover cases when the procfs library encounters something
unexpected and would otherwise panic.  Every instance of `InternalError`
should be considered a bug that should be reported

Closes #44

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,9 @@ categories = ["os::unix-apis", "filesystem"]
 license = "MIT OR Apache-2.0"
 edition = "2018"
 
+[features]
+default = []
+
 [dependencies]
 libc = "0.2"
 bitflags = "1"
@@ -18,6 +21,7 @@ chrono = "0.4"
 byteorder = {version="1", features=["i128"]}
 hex = "0.3"
 libflate = "0.1"
+backtrace = { version = "0.3", optional = true }
 
 [dev-dependencies]
 failure = "0.1.5"
--- a/examples/netstat.rs
+++ b/examples/netstat.rs
@@ -8,7 +8,7 @@ use std::collections::HashMap;
 
 fn main() {
     // get all processes
-    let all_procs = procfs::all_processes();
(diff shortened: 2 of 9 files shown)


eminence:
Things I want to do before publishing a new version with this change:

- [ ] Do another code review, in particular look for panicing functions that aren't named `unwrap()`
- [ ] Review test cases and see if any functions aren't covered (maybe this is a good time to investigate some type of code-coverage robot)
- [ ] Tweak the Debug/Display impls for `InternalError` to make sure they show useful into when displayed
- [ ] Think about adding an optional backtrace to `InternalError` (probably off by default, but could be enabled with a feature flag)

idanski:
@eminence Nice! I think maybe the only input I have at the moment is that `InternalError(InternalError)` maybe a bit confusing to newcomers?

The other thing is that maybe ProcError can implement the Error trait but that is maybe unrelated to this effort :)

Oh, and I heard about this: https://github.com/japaric/panic-never 
Haven't used it, but it may help?

Thanks!

eminence:
Thanks for taking a look!  I'm fairly happy with this right now, so I'm going to to merge it.  

If needed, we can improve the documentation to help people understand what to do with an `ProcError::InternalError(InternalError)`.

> The other thing is that maybe ProcError can implement the Error trait 

[Already done](https://docs.rs/procfs/0.5.4/procfs/enum.ProcError.html#implementations)

> Oh, and I heard about this: https://github.com/japaric/panic-never

Looks neat, it's worth taking a look.  Also maybe https://github.com/dtolnay/no-panic

PR #56, opened 2019-12-16
Introduce Process::new_with_root() constructor

edigaryev:
This allows querying information about processes on systems where procfs is mounted to directories other than /proc.

### DIFF
--- a/src/process.rs
+++ b/src/process.rs
@@ -1,9 +1,12 @@
 //! Functions and structs related to process information
 //!
-//! The primary source of data for functions in this module is the files in a /proc/<pid>/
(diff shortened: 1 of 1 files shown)


eminence:
Looks good, thanks!  Out of curiosity, what kind of system are you on that doesn't have procfs mounted on `/proc` ?

edigaryev:
> Looks good, thanks! Out of curiosity, what kind of system are you on that doesn't have procfs mounted on `/proc` ?

Containers.

eminence:
Both of your changes are now published as `v0.7.6`.  Thanks!

PR #58, opened 2019-12-19
Fix Stat's starttime, guest_time and cguest_time field integer types

edigaryev:
[proc(5) manual page](http://man7.org/linux/man-pages/man5/proc.5.html) lists the following data types for `/proc/[pid]/stat`:

* `%llu` for starttime, which is unsigned
* `%lu` for guest_time, which is 64-bit on [LP64](https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models)
* `%ld` for cguest_time, which is signed and 64-bit on [LP64](https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models)


eminence:
Thanks 👍

PR #59, opened 2019-12-19
Implement Process.root() method to retrieve current root directory

edigaryev:
This PR also makes `Process.cwd()` and `Process.exe()` documentation more consistent.


eminence:
Thanks 👍

PR #61, opened 2019-12-23
UDP sockets have a state too

edigaryev:
Without the state there's no way to distinguish between UDP sockets in LISTEN and ESTABLISHED states.

One can reproduce this by starting `nc -u 127.0.0.1 1234` and examining the `netstat -anu` output.

### DIFF
--- a/src/net.rs
+++ b/src/net.rs
@@ -108,6 +108,29 @@ impl TcpState {
     }
 }
 
+#[derive(Debug, PartialEq)]
+pub enum UdpState {
+    Established = 1,
+    Close = 7,
+}
+
+impl UdpState {
+    pub fn from_u8(num: u8) -> Option<UdpState> {
+        match num {
+            0x01 => Some(UdpState::Established),
+            0x07 => Some(UdpState::Close),
+            _ => None,
+        }
+    }
+
+    pub fn to_u8(&self) -> u8 {
+        match self {
+            UdpState::Established => 0x01,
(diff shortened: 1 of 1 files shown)


src/net.rs
```diff
@@ -107,6 +107,7 @@ pub struct TcpNetEntry {
 pub struct UdpNetEntry {
     pub local_address: SocketAddr,
     pub remote_address: SocketAddr,
+    pub state: TcpState,
```
eminence: I think I would prefer to have a new `UdpState` enum for this (since a UDP socket can't be in all of the same states that a TCP socket can)

eminence: (I realize that internally the kernel uses the same state type values as TCP connections, but having a new `UdpState` would reduce confusion I think)

edigaryev: You mean something like this?

```rust
enum UdpState {
    Established = 1,
    Close = 7,
}
```

eminence: Yeah, exactly

edigaryev: See 8baacbe.

edigaryev: See 8a55ab4f0cac1d36077ba88149441d7b745bc222.

eminence:
Looks great, thank you!

eminence:
This inspired me to look at unix sockets, and they too have a state.  I've added this in 48b1ee1e938c537e261e9a5a856e0020889e4c76

PR #64, opened 2020-01-23
Add boot_time cache to improve performance

dalance:
This PR improves `Stat::starttime()` performance.
Currently `starttime()` is slow because it internally calls `boot_time()` each time.
I think the result of `boot_time()` can be cached because it is not changed through program execution.

I tried the following benchmark:

```rust
#[bench]
fn bench_starttime(b: &mut test::Bencher) {
    b.iter(|| {
        for p in process::all_processes().unwrap() {
            let _ = p.stat.starttime();
        }
    });
}
```

The result is below:

* Current master

```
test tests::bench_starttime ... bench: 695,472,752 ns/iter (+/- 35,133,377)
```

* This PR applied

```
test tests::bench_starttime ... bench:   7,444,747 ns/iter (+/- 1,975,527)
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -641,8 +641,20 @@ pub fn boot_time() -> ProcResult<DateTime<Local>> {
     doc = "See also [boot_time()] to get the boot time as a `DateTime`"
 )]
 pub fn boot_time_secs() -> ProcResult<u64> {
-    let stat = KernelStats::new()?;
-    Ok(stat.btime)
+    BOOT_TIME.with(|x| {
+        let mut btime = x.borrow_mut();
+        if let Some(btime) = *btime {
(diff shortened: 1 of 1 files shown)


eminence:
On my system, the current master branch is about an order of magnitude faster than your system, which is interesting:

    test tests::bench_starttime ... bench:  36,540,102 ns/iter (+/- 2,651,593)

And with this patch applied, it's quite a bit faster:

    test tests::bench_starttime ... bench:   8,893,658 ns/iter (+/- 917,462)  

Looks like a win! 

By my analysis, the call to `borrow_mut` will never panic because this is a thread-local variable that can never be shared between threads (because `RefCell` is `!Send`), so it will never be borrowed simultaneously.

PR #67, opened 2020-03-13
Fix incorrect IPv6 address parsing

edigaryev:
If you look at [how Linux formats the /proc/net/tcp6 internally](https://elixir.bootlin.com/linux/v4.19/ident/get_tcp6_sock), you can notice that the address is actually consists of four 32-bit integers and they are emitted using system endianess.

The [same applies to /proc/net/udp6](https://elixir.bootlin.com/linux/v4.19/ident/__ip6_dgram_sock_seq_show).

Unwrap each of the integer separately and cut them into [hextets](https://en.wikipedia.org/wiki/Hextet) for Ipv6Addr::new()'s pleasure.

Also bring the tests in touch with reality.

### DIFF
--- a/src/net.rs
+++ b/src/net.rs
@@ -49,7 +49,7 @@ use crate::ProcResult;
 use std::collections::HashMap;
 
 use crate::FileWrapper;
-use byteorder::{ByteOrder, NetworkEndian};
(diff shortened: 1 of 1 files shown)


eminence:
great catch.  Thank you!

eminence:
This fix has been published as v0.7.8

PR #83, opened 2020-06-17
fix typo: hugetblpages -> hugetlbpages

liubin:
`hugetblpages` may be a typo of `hugetlbpages`


eminence:
Agreed!  Good catch, thank you.

PR #84, opened 2020-06-24
adds pid_max

wookietreiber:
This should be pretty self-explanatory, available since 2.5.34. Don't know if that requires checking, but I guess not.

Let me know if I should add a `pub use` in any of `lib.rs` or `sys/mod.rs`, not sure if that's needed/wanted.

---

I recently needed this for checking user input of a CLI tool that uses PIDs as arguments, see example in the docs.

### DIFF
--- a/src/sys/kernel/mod.rs
+++ b/src/sys/kernel/mod.rs
@@ -119,6 +119,27 @@ impl cmp::PartialOrd for Version {
     }
 }
 
+/// Returns the maximum process ID number.
+///
+/// This is taken from `/proc/sys/kernel/pid_max`.
+///
+/// # Example
+///
+/// ```
+/// let pid_max = procfs::sys::kernel::pid_max().unwrap();
+///
+/// let pid = 42; // e.g. from user input, CLI args, etc.
+///
+/// if pid > pid_max {
+///     eprintln!("bad process ID: {}", pid)
+/// } else {
+///     println!("good process ID: {}", pid);
+/// }
+/// ```
+pub fn pid_max() -> ProcResult<i32> {
+    read_value("/proc/sys/kernel/pid_max")
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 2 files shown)


src/sys/kernel/mod.rs
```diff
@@ -119,6 +119,27 @@ impl cmp::PartialOrd for Version {
...
+/// This is taken from `/proc/sys/kernel/pid_max`.
+///
+/// # Example
+///
+/// ```
+/// let pid_max = procfs::sys::kernel::pid_max().unwrap();
+///
+/// let pid = 42; // e.g. from user input, CLI args, etc.
+///
+/// if pid > pid_max {
```
eminence, line 133: This if statement looks backwards to me

src/sys/kernel/mod.rs
```diff
@@ -119,6 +119,27 @@ impl cmp::PartialOrd for Version {
...
+///     println!("good process ID: {}", pid);
+/// } else {
+///     eprintln!("bad process ID: {}", pid)
+/// }
+/// ```
+pub fn pid_max() -> ProcResult<libc::pid_t> {
```
eminence, line 139: Can you change the return type to be an `i32` instead of `libc::pid_t`?   I know that `pid_t` is actually the correct return type, but unfortunately I use `i32` everywhere else (for example [here](https://docs.rs/procfs/0.7.9/procfs/process/struct.Process.html#structfield.pid), [here](https://docs.rs/procfs/0.7.9/procfs/process/struct.Stat.html#structfield.pid), and [here](https://docs.rs/procfs/0.7.9/procfs/process/struct.MountInfo.html#structfield.pid)).   Eventually I might change everything into a `pid_t`, but until then I'd prefer all the pid types to be consistent.

wookietreiber: Done.

src/sys/kernel/mod.rs
wookietreiber: Oops, thanks for spotting, changed.

eminence:
Thanks!

PR #86, opened 2020-07-05
Add the ability to list all tasks inside a process

quodlibetor:


eminence:
This is looking good to me, thank you!  There's only 1 topic I want to quickly discuss:  what's your thought on the laziness of the TaskIter?  Do you think there is a potential hazard in someone getting a TaskIter, but then iterating over it slowly enough that the task would have disappeared by the time the iterator got to it?  Obviously your code handles this by returning a ProcResult.  Do you see any advantage in eagerly building the tasks when calling `tasks()` and maybe returning a `ProcResult<Vec<Task>>`?  

(To be clear: I don't have any objection to your implementation, I just wanted to open the point for discussion).

Also, could you please update `support.md` ?

quodlibetor:
>  Do you think there is a potential hazard in someone getting a TaskIter, but then iterating over it slowly enough that the task would have disappeared by the time the iterator got to it?

Absolutely, this is a risk. AFAICT though the exact same risk occurs even if we return a `ProcResult<Vec<Task>>`, really we'd need to return `ProcResult<Vec<ProcResult<Task>>>` because threads can end in between when we start iterating and fill the things.

With the current lazy implementation it is strictly more flexible, I think:

* Theoretically doing what the test does (just immediately collecting) should work out to exactly the same code as returning the Vec.
* We don't need to make a policy decision about if whether we should abort the vec creation depending on the kind of error we run into -- we push the responsibility of deciding that back onto the user, without requiring that we build the entire vec if there are errors
* Optimization-wise, this allows users to not actually create the vec, or to re-use a vec (via `clear`/`from_iter`)if they want. (My use case doesn't require holding onto everything, just extracting a couple properties, and losing some data is fine).

Actually, I noticed your "future breaking changes" issue after I made this PR, and I'm tempted to make the `Task` struct even more lazy, and not include the `stat` property and just adding a method. OTOH that would make it unnecessarily distinct from the Process, which seems like an unnecessary burden.

> Also, could you please update support.md ?

Happy to!

eminence:
Just to expand my thinking for posterity: if someone did something like this:

```rust
for task in p.tasks()? {
    do_slow_operation(task?);
}
```

They may see more errors (missing tasks) than if the results were collected immediately (which still has the possibility for tasks to end while iterating the results of readdir, but it's just less likely).  Anyway, I don't think it's a big enough deal to change the design.


>  I'm tempted to make the Task struct even more lazy, and not include the stat property and just adding a method

I'd be OK with this.  I don't particularly mind the lack of symmetry if `Process` contains a `stat` field, but `Task` doesn't.   I'll leave it up to you :)

quodlibetor:
Yeah I agree with your concerns. Perhaps it would be helped if I documented the behavior and that if you care a lot about getting the most possible information you should collect into some container immediately? FWIW I'm collecting into a dict, aggregated by `comm`, which is why I don't particularly want the `Vec` to just throw it out.

I'm honestly not sure what happens with the directory iterator if a new thread is started after iteration starts, either, I might check it.

>  I don't particularly mind the lack of symmetry if Process contains a stat field, but Task doesn't

I'll convert `stat` into a method!

quodlibetor:
So I ran a bunch of different tests (in the travis CI log) and it seems like:

* Tasks are not eagerly cleaned up stopping the thread and waiting for it doesn't guarantee that we can't find it in the list
* Tasks that are started (lexically...) after the iterator do reliably show up in the task list

So being lazy seems likely to provide strictly more tasks than being eager, but I could see arguments that either is more "correct". Still the fact that the iterator is both more flexible and possibly more efficient encourage me to keep it as an iterator.

eminence:
Ok, that sounds great, thanks for running those tests!  Let's merge this

quodlibetor:
Thanks! Is there anything I can do to help get this released? procfs is a few levels deep in my dependency hierarchy and patching is a little complex.

eminence:
I'll get this published on crates.io later today or tomorrow (i'll ping you here when that happens).

quodlibetor:
Thanks so much!

eminence:
Published as `v0.8.0` 🚀

PR #93, opened 2020-09-03
Add a method to get a task for main thread

nbaksalyar:
While this method is not strictly required (we can always get the main task by filtering `TasksIter` as `process.tasks().filter(...)`), I think it improves the API ergonomics for cases when we care about the main thread only and makes it more simple because the main task ID is always equal to the process ID.

### DIFF
--- a/src/process/mod.rs
+++ b/src/process/mod.rs
@@ -888,6 +888,11 @@ impl Process {
         StatM::from_reader(file)
     }
 
+    /// Return a task for the main thread of this process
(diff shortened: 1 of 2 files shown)


eminence:
Looks good to me, thanks!   I'll get this published to crates.io sometime in the next day or two

nbaksalyar:
Thank you!

eminence:
This is now published as `v0.8.1`

PR #95, opened 2020-09-23
Add Android support

fabricedesre:
This adds support for android targets. Changes needed are:
- use the android specific version of the `MetadataExt` trait.
- use `libc::mode_t` instead of hardcoding the type for file modes (it's a u16 on 32bit Android targets).

### DIFF
--- a/src/process/mod.rs
+++ b/src/process/mod.rs
@@ -58,8 +58,10 @@ use crate::from_iter;
 use std::ffi::OsString;
 use std::fs;
 use std::io::{self, Read};
-#[cfg(unix)]
+#[cfg(all(unix, not(target_os = "android")))]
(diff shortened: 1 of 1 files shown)


eminence:
LGTM, thanks!

I'm not an expert in android development, so a quick question for you:  do you have any recommendations on testing android in CI?   Is running `cargo check --target aarch64-linux-android` sufficient?  Or are there other targets that should be also checked?

fabricedesre:
Yes, running `cargo check --target aarch64-linux-android` is a good way to verify the Rust parts, while not having to deal with the link phase which is more annoying because you need an android sysroot.

eminence:
Ok great.  I'll add that later.  Thank you!

eminence:

PR #97, opened 2020-10-03
Add github action workflow check that enforces 'cargo fmt'

Nukesor:
Hey there :)

This PR adds a `cargo fmt` check to the `rust` Github action.
That should enforce/remind people to use `cargo fmt` in their PRs.

I'm also thinking about adding another PR, which resolves all open `cargo clippy` issues and adds a cargo clippy check to the action. If you don't mind, I'll start working on that as well momentarily.

This also adds a rule, which actually runs the checks on PRs.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -36,6 +36,9 @@ jobs:
       if: matrix.toolchain == '1.33.0'
       run: cargo +${{ matrix.toolchain }} doc
 
+    - name: Run cargo fmt
+      if: matrix.toolchain == 'stable'
+      run: cargo +${{ matrix.toolchain }} fmt -- --check
 
   android-check:
     strategy:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,8 +1,12 @@
 name: Rust
 
-on: [push]
-
-
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - 'v*.*.*'
+  pull_request:
 
 jobs:
   build:
--- a/src/process/mod.rs
+++ b/src/process/mod.rs
@@ -58,10 +58,10 @@ use crate::from_iter;
 use std::ffi::OsString;
 use std::fs;
 use std::io::{self, Read};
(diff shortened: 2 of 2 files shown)


Nukesor:
I'm not exactly sure why this test is failing. This doesn't look related to any changes I made. The other PR that builds upon this one looks like it's working.

Maybe you have some non-deterministically failing tests?

eminence:
Yeah, some of the tests are a little racey (a few milliseconds go by between when we get a process listing and when we try to get info about it -- in that time, the process might have exited).  I've re-run the tests and they are passing now.

This same comment applies to both of your pull requests:  I don't really have an objection to adding clippy or rustfmt to CI (as you'll see several commits in the git history were I apply clippy or fmt fixes), but I'm a little hesitant to block PRs on them.  It's a bit of a barrier/annoyance that I'm OK dealing with myself, but I don't necessarily want all contributors to deal with it if they don't want to.

Can we try to mark the rustfmt and clippy jobs as ["continue-on-error"](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error)  (and then introduce a "fake" clippy or fmt error to see what it looks like)?    I've not used this github action setting before, but I'm hoping it might be like gitlab's "allow-failure" where it turns the job a yellow color.

Nukesor:
I looked around and those configs aren't an option yet.

This allows workflows to not fail, when some steps fail, but there's no indication that anything went wrong.
https://github.com/actions/toolkit/issues/399

A workaround would be to set the outcome status manually, but this isn't possible yet either -.-
https://github.com/actions/runner/issues/662

Github actions really feels like a beta CI product.

eminence:
Thanks for researching this a bit.  It's a bummer that github doesn't support this, though hopefully one day they will!  Something like Gitlab's "allow_failure: true" would fit well here, I think.

So regarding these two pull requests, let's close them, since github actions can't really do what we want.

If you wanted to open another PR with some clippy/fmt fixes, I'd be happy to accept that :)

Issue #69, opened 2020-03-27
Future breaking API changes


eminence:
This issue is collecting some ideas for breaking API changes that we might want to make in the future

* Convert some fields from `Option<T>` to `T` if they were added in very old kernels (#68)
* Remove `stat` and `pid` fields from the `Process` struct to support cases when you don't need this information (#60)
* Remove some deprecated methods (#52)
* Returning ticks in CpuTime (#76)

madjokr:
Hi,
I just started using this library. First off, it's a fantastic library. I am using this library to build a service that's going to be deployed around 1000+ instances. My app uses PID field to monitor the process. If you are removing this option, will there an alternative way to find process PID through this library? Sorry for sending a message this late.

eminence:
Hello!  I'm glad to hear that this library is useful to you!

The proposal in #60 is to just remove the `pid` *field*.  So you can still get access to the PID information via the `stat()` method.  For example, to get the PID of the current process:

```rust
let me = procfs::process::Process::myself().unwrap();
let my_pid = me.stat().unwrap().pid;
```

(This code works today, too).

Does that answer your question?

lparcq:
Hello,
The time returned in procfs::CpuTime should be a number of ticks in u64 as it is for processes in procfs::process::Stat. It's better to be able to get the values as returned by the kernel.

Nukesor:
I'm just in the process of removing all clippy warnings and adding clippy checks.

This involves replacing a lot of manual prefix removals by the new `strip_prefix` method.
Since this method has been added in Rust `1.45`, this would certainly remove support for `1.33` and `1.38` versions.

I guess this could be considered a breaking change.

An alternative would be to simply ignore these clippy warnings.

eminence:
Hi @uggla :wave:

The short answer is:  the API can still change in incompatible ways in a future 0.10.x release.  I don't consider the API stable enough for a 1.0 release yet.  (In fact there's an open draft PR that is proposing a breaking change).  However in practice, the API breaking changes tend to be fairly small, and not that often.

I saw the comment you made in the scaphandre repo, so I'll also make a comment over there to talk about any specific issues related to your project.

eminence:
This issue hasn't been updated in a long while, so I'm going to close it.  We've been doing breaking changes in new minor versions, and this has been working well for us for a while.  Thanks all

PR #98, opened 2020-10-03
Fix clippy and rustfmt issues

Nukesor:
I just took the opportunity and started to hack around a little :)

This PR is build upon the other `cargo fmt` branch.

This probably needs some discussion as mentioned in issue #69.
There are two `strip_prefix` commits, which replace manual string prefix removal by the newly added `strip_prefix` method and one commit which renames the deprecated `intra_doc_link_resolution_failure`.

Since the latest feature has only just been added in Rust `1.45`, this will drop support for rust `1.38` and `1.33`.
I'm not sure what your policy for support of old rust versions is. In case you don't want to drop support for these versions, I would disable all clippy rules for all breaking changes and reintroduce these targets to the CI rules.

Anyway, I hope you would like to see clippy being added to this project's CI

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,6 @@
 // Don't throw clippy warnings for manual string stripping.
 // The suggested fix with `strip_prefix` removes support for Rust 1.33 and 1.38
+#![allow(clippy::unknown_clippy_lints)]
 #![allow(clippy::manual_strip)]
 // Don't throw rustc lint warnings for the deprecated name `intra_doc_link_resolution_failure`.
 // The suggested rename to `broken_intra_doc_links` removes support for Rust 1.33 and 1.38.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,3 +1,9 @@
+// Don't throw clippy warnings for manual string stripping.
+// The suggested fix with `strip_prefix` removes support for Rust 1.33 and 1.38
+#![allow(clippy::manual_strip)]
+// Don't throw rustc lint warnings for the deprecated name `intra_doc_link_resolution_failure`.
+// The suggested rename to `broken_intra_doc_links` removes support for Rust 1.33 and 1.38.
+#![allow(renamed_and_removed_lints)]
 #![deny(intra_doc_link_resolution_failure)]
 //! This crate provides to an interface into the linux `procfs` filesystem, usually mounted at
 //! `/proc`.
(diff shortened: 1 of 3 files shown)


eminence:
HI! Thanks for tackling the clippy issues.  While code with `strip_prefix` is clearly much nicer, I'm not willing to drop support for 1.33 and 1.38

Nukesor:
Hey :)

I removed all changes to the Github actions file, and disabled all warnings for stuff which would break backward compatibility.
For each exception, a comment has been added which explains why this exception exists.

I guess the PR is ready for the next review round

src/lib.rs
```diff
@@ -606,7 +612,7 @@ pub fn ticks_per_second() -> std::io::Result<i64> {
     if cfg!(unix) {
         match unsafe { sysconf(_SC_CLK_TCK) } {
             -1 => Err(std::io::Error::last_os_error()),
-            x => Ok(x.into()),
+            x => Ok(x),
```
eminence, line 615: Can you leave this as `x.into()`?  Removing this will break the build on 32-bit, since on 32-bit linux, `sysconf` returns an `i32`.  So the `.x.into()` will widen this to an `i64`.  Same comment applies below to line 671.

src/sys/fs/binfmt_misc.rs
```diff
@@ -6,14 +6,6 @@ pub fn enabled() -> ProcResult<bool> {
     Ok(val == "enabled")
 }
 
-struct BinFmtEntry {
```
eminence, line 9: I know this struct is private and unused, but it'll soon be exposed.  Can you leave it in?

eminence (changes requested):
Looks good, with just 2 requested changes.

eminence (approved):

eminence:
This is minor, but since we're in this part of the code..... What do you think about adding/allowing `clippy::unknown_clippy_lints` to quell this warning:

```
warning: unknown clippy lint: clippy::manual_strip
 --> src/net.rs:3:10
  |
3 | #![allow(clippy::manual_strip)]
  |          ^^^^^^^^^^^^^^^^^^^^ help: did you mean: `clippy::manual_swap`
  |
  = note: `#[warn(clippy::unknown_clippy_lints)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unknown_clippy_lints

```

Nukesor:
I added the unknown_clippy_lints rule :)
This should remove that clippy warning in older versions

eminence:
Thank you!

PR #105, opened 2020-10-23
Introduce fd_count() in addition to fd()

bobrik:
Prometheus crate uses `fd()` to return the number of open file descriptors. Currently it takes 500ms to count 80k open files in an uncontended case, with this method we can cut this down to just 100ms by not resolving details of every open file descriptor.

See: https://github.com/tikv/rust-prometheus/blob/512e8ed44c0fbbcef/src/process_collector.rs#L134-L137

### DIFF
--- a/src/process/mod.rs
+++ b/src/process/mod.rs
@@ -743,6 +743,13 @@ impl Process {
         Ok(vec)
     }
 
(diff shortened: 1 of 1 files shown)


eminence:
I like it, thanks!

eminence:

PR #111, opened 2020-11-16
Add support for /proc/<pid>/schedstat

zpp0:


eminence:
Thanks for this new function, this looks good.

I have this `schedstat` file in my procfs filesystem, but it not documented in my `proc(5)` man page.  Do you know if this file is documented anywhere (other than the kernel source?)

zpp0:
Thanks for the review. The file is roughly documented by https://www.kernel.org/doc/html/latest/scheduler/sched-stats.html#proc-pid-schedstat 
Measurement units of the fields aren't documented, as far as I know - I had to look at the kernel code and test.
The commit message of 2e11080 has a link to the kernel documentation, just in case.

eminence:
OK, great.  I added a small addition in ea3e8a999e2068367973c43cc6efe7b8b9daa7f5 to say that `CONFIG_SCHED_INFO` is needed.

PR #112, opened 2020-12-01
Add support for /proc/<pid>/smaps

RealKC:
I wasn't exactly sure how to test this, as `/proc/<pid>/smaps` is only available if the kernel was built with the  CONFIG_PROC_PAGE_MONITOR config option, though both my Arch Linux kernel and Raspbian kernel seem to have been built with it, so it is probably a common option? I'd like to hear your opinion on this first however.

### DIFF
--- a/src/process/mod.rs
+++ b/src/process/mod.rs
@@ -545,7 +545,7 @@ impl MemoryMap {
 /// Represents the information about a specific mapping as presented in /proc/<pid>/smaps
 ///
 /// To construct this structure, see [Process::smaps()]
-#[derive(Default)]
+#[derive(Default, Debug)]
 pub struct MemoryMapData {
     /// Key-Value pairs that may represent statistics about memory usage, or other interesting things,
     /// such a "ProtectionKey"(if you're on X86 and that kernel config option was specified).
--- a/src/process/mod.rs
+++ b/src/process/mod.rs
@@ -190,6 +190,123 @@ bitflags! {
     }
 }
 
+bitflags! {
+    /// Represents the kernel flags associated with the virtual memory area.
+    /// The names of these flags are just those you'll find in the man page, but in upper case.
+    pub struct VmFlags: u32 {
+        /// Invalid flags
+        const INVALID = 0;
+        /// Readable
+        const RD = 1 << 0;
+        /// Writable
+        const WR = 1 << 1;
+        /// Executable
+        const EX = 1 << 2;
+        /// Shared
+        const SH = 1 << 3;
+        /// May read
+        const MR = 1 << 4;
+        /// May write
+        const MW = 1 << 5;
+        /// May execute
+        const ME = 1 << 6;
+        /// May share
+        const MS = 1 << 7;
+        /// Stack segment grows down
+        const GD = 1 << 8;
+        /// Pure PFN range
+        const PF = 1 << 9;
+        /// Disable write to the mapped file
+        const DW = 1 << 10;
+        /// Pages are locked in memory
+        const LO = 1 << 11;
+        /// Memory mapped I/O area
+        const IO = 1 << 12;
+        /// Sequential read advise provided
+        const SR = 1 << 13;
+        /// Random read provided
+        const RR = 1 << 14;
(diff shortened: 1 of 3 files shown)


src/process/mod.rs
```diff
@@ -388,6 +510,54 @@ pub struct MemoryMap {
...
+/// Represents the information about a specific mapping as presented in /proc/<pid>/smaps
+///
+/// To construct this structure, see [Process::smaps()]
+#[derive(Default)]
+pub struct MemoryMapData {
+    /// Key-Value pairs that may represent statistics about memory usage, or other interesting things,
+    /// such a "ProtectionKey"(if you're on X86 and that kernel config option was specified).
+    ///
+    /// Check your manpage for more information
+    pub key_value_pairs: HashMap<String, u64>,
```
RealKC, line 554: I don't like this name, however I am not sure what a better name would be. Any suggestions?

eminence: What about just calling this `map`  (since it is a HashMap after all)?   Or maybe `data`?

RealKC: I went with `map` since I liked that name more, but I don't have a particularly strong opinion on this.

src/process/mod.rs
```diff
@@ -719,12 +872,79 @@ impl Process {
...
+                    let key = parts.next();
+                    let value = parts.next();
+
+                    if let (Some(k), Some(v)) = (key, value) {
+                        // While most entries do have one, not all of them do.
+                        let size_suffix = parts.next();
+
+                        // Limited poking at /proc/<pid>/smaps and then checking if "MB", "GB", and "TB" appear in the C file that is
+                        // supposedly responsible for creating smaps, has lead me to believe that the only size suffixes we'll ever encounter
+                        // "kB", which is most likely kibibytes. Actually checking if the size suffix is any of the above is a way to
```
eminence, line 928: Looking at [fs/proc/task_mmu.c](https://github.com/torvalds/linux/blob/master/fs/proc/task_mmu.c#L775) concurs with your comment that "kB" is the only used used here, so I'm OK with the code as you've written it.

eminence:
This is looking great, thank you!   Two quick comments:

1. Can you please add a test?   Just something simple that will validate that the parsing code doesn't explode.  Since this file isn't always available on all kernels, one way to write this test would be something like:

```rust
#[test]
fn test_smaps() {
    let me = Process::myself().unwrap();
    let smaps = match me.smaps() {
        Ok(x) => {x}
        Err(ProcError::NotFound(_)) => {
            // ignored because not all kernerls have smaps
            return;
        }
        Err(e) => panic!("{}", e)
    };

    println!("{:#?}", smaps);
}
```


2. The [maps()](https://docs.rs/procfs/0.9.0/procfs/process/struct.Process.html#method.maps) method returns a `Vec<MemoryMap>`  while smaps() returns `HashMap<MemoryMap, MemoryMapData>`.  This inconsistency isn't fatal, but let's take a moment to think about it.   I *think* (but I'm not sure) that both /proc/pid/maps and /proc/pid/smaps are ordered by address.  I suspect this ultimately all that useful (since my guess is that most people will be searching for thigns in maps/smaps, not displaying them in an ordered list), but it does make me gravitate to a sorted data structure.   Extracting the original ordering from the HashMap wouldn't be hard, though.   Also, it doesn't seem likely to me that someone would have a `MemoryMap` and need to look up its corresponding `MemoryMapData`.  More likely they'd iterate over both at the same time.  What do you think about having smaps() return something like `Vec<(MemoryMap, MemoryMapData)>` ?

RealKC:
Re 1: Will do.

Re 2: You bring good points here, so I think switching to returning `Vec<(MemoryMap, MemoryMapData)>` would be a good idea. Should I then remove the stuff I added so `MemoryMap` is hashable?

eminence:
Eh, I think it's fine for `MemoryMap` to stay hashable.  I don't have a strong feeling either way.

eminence (approved):

eminence:
Thanks!

eminence:
This is now published on crates.io as v0.9.1

PR #113, opened 2020-12-04
Fix build

RealKC:
strip_suffix was introduced in 1.45 while our MSRV is 1.34

-------

Sorry for that, not yet used to working with older compiler versions. Maybe it would be a good idea to run your checks on PRs as well so that using features that are too new will fail early?

### DIFF
--- a/src/process/mod.rs
+++ b/src/process/mod.rs
@@ -934,7 +934,7 @@ impl Process {
                         })?;
 
                         // This ignores the case when our Key: Value pairs are really Key Value pairs. Is this a good idea?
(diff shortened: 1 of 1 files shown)


eminence:
I'm fairly new to github actions, and I'm not sure why they aren't running for your pull requests.  I'll do some research (though if anyone reading this knows, please let me know!)

Thanks for the quick fix!

RealKC:
Changing the `on: [push]` line from your GitHub Action to `on: [push, pull_request]` should make GitHub Actions run on pushes and pull requests I believe

eminence:
Ahh, perfect.    Want to open a PR for that to test? 😃

PR #115, opened 2020-12-07
Add support for a bunch of low hanging fruits

RealKC:
Most of these files didn't require any complicated parsing, if any at all.

Also I ended up fixing a bug, I believe. (Not sure how that didn't come up earlier?)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -55,7 +55,6 @@ use libc::pid_t;
 use libc::sysconf;
 use libc::{_SC_CLK_TCK, _SC_PAGESIZE};
 
-use std::ffi::CStr;
 use std::fmt;
 use std::fs::File;
 use std::io::{self, BufRead, BufReader, Read, Write};
@@ -64,6 +63,7 @@ use std::os::raw::c_char;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::{collections::HashMap, time::Duration};
+use std::{ffi::CStr, fs::OpenOptions};
 
 #[cfg(feature = "chrono")]
 use chrono::{DateTime, Local};
@@ -228,7 +228,7 @@ pub(crate) fn read_file<P: AsRef<Path>>(path: P) -> ProcResult<String> {
 }
 
 pub(crate) fn write_file<P: AsRef<Path>, T: AsRef<[u8]>>(path: P, buf: T) -> ProcResult<()> {
-    let mut f = File::open(path)?;
+    let mut f = OpenOptions::new().write(true).open(path)?;
     f.write_all(buf.as_ref())?;
     Ok(())
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1090,6 +1090,19 @@ pub fn modules() -> ProcResult<HashMap<String, KernelModule>> {
     Ok(map)
 }
 
+/// Get a list of the arguments passed to the Linux kernel at boot time
+///
+/// This corresponds to the data in `/proc/cmdline`
+pub fn cmdline() -> ProcResult<Vec<String>> {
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -228,7 +228,7 @@ pub(crate) fn read_file<P: AsRef<Path>>(path: P) -> ProcResult<String> {
 }
 
 pub(crate) fn write_file<P: AsRef<Path>, T: AsRef<[u8]>>(path: P, buf: T) -> ProcResult<()> {
-    let mut f = File::open(path)?;
+    let mut f = OpenOptions::new().write(true).open(path)?;
```
eminence, line 231: Hah, that's an embarrassing bug.   I guess no one uses the write functionality.  I didn't personally add it, so I don't have a good sense of the use cases for this write functionality.  Also, clearly we have no automated tests for this (I"m going to open a new issue on that topic shortly)

RealKC: In general, I'd say that testing the write functionality is hard as writing to the writable files under /proc has system-wide effects. What I think could be done is running tests under a chroot with a fake /proc, could help make us more confident in the read tests too, but I'm not sure it'd be that easy to setup(not sure docker(or a similar container tool) can help here, unless it'd allow us to make fake files under /proc)

eminence:
Thanks for this PR, sorry I haven't gotten a chance to review it yet (things are very busy at $DAYJOB).   Just letting you know I haven't forgotten about this

RealKC:
No worries, take your time!

src/sys/kernel/mod.rs
```diff
@@ -126,6 +129,188 @@ pub fn pid_max() -> ProcResult<i32> {
...
+            semmni,
+        })
+    }
+}
+
+impl FromStr for SemaphoreLimits {
+    type Err = &'static str;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        SemaphoreLimits::from_str(s)
```
eminence, line 176: Instead of forwarding `FromStr::from_str` to `SemaphoreLimits::from_str`, would it be simpler to just have the actual implementation in this method?

src/lib.rs
eminence: Running tests under a chroot with a fake /proc is a clever idea, I like it!  Initially I had in mind a system where the tests are run under some qemu/kvm VM, which would let us write these values without worrying about impacting a real running system.  Running an actual VM is rather heavyweight, but it might be OK since I think the tests would be run perhaps once per release

eminence:
This new stuff looks great, thank you!

A quick comment (to no one in particular) -- it may seem like these functions aren't providing a lot of value, since they are so small and in some cases the data is very easy to parse.  But in my mind the real value here is the documentation that describes what each function/field is for (which isn't always easy to remember or lookup).  I am quite happy that @RealKC did a nice job adding these documentations.  If anyone else reading this wants to make a PR that is "only documentation" please know that this is valuable and really important too!

src/lib.rs
RealKC: I think that running a VM could work as well, but I'm not sure how good the tools to instrument VMs for this kind of thing are, and I'm not sure everything would be easily inspectable that way.

I think having a deeper look into both approaches would be a good idea though.

src/sys/kernel/mod.rs
RealKC: I think I followed the example of another struct while doing this, though I wouldn't be opposed to just inlining the implementation in the trait impl

eminence: ok cool.  it's a super minor thing, so let's get this merged and someone can clean this up later 🚀

PR #116, opened 2021-01-13
Override default cpuinfo source

rjbergTU:
Overriding the source of cpuinfo information can help in some situations. For example, parsing a file that comes from another system.


eminence:
Thanks. I'll get this published on crates.io soon

PR #119, opened 2021-01-28
Change MSRV when using the `backtrace` feature to 1.42

eminence:


eminence:
Hmm, why isn't github actions running on this PR...

PR #123, opened 2021-03-11
Create `FDInfo` from raw fd and pid

saruman9:
Sometimes it's needed for working with single high level `FDInfo` not raw fd, but getting all fds from `Process` is not suitable.

### DIFF
--- a/src/process/mod.rs
+++ b/src/process/mod.rs
@@ -690,7 +690,7 @@ impl FDInfo {
     pub fn from_raw_fd_with_root(root: impl AsRef<Path>, pid: pid_t, raw_fd: i32) -> ProcResult<Self> {
         let path = root.as_ref().join(pid.to_string()).join("fd").join(raw_fd.to_string());
         let link = wrap_io_error!(path, read_link(&path))?;
-        let md = wrap_io_error!(path, path.metadata())?;
+        let md = wrap_io_error!(path, path.symlink_metadata())?;
         let link_os: &OsStr = link.as_ref();
         Ok(Self {
             fd: raw_fd as u32,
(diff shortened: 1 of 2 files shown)


eminence:
Hey, thanks for the PR!  I'm just leaving a note here to say that I haven't forgotten about this, and will review this soon!

eminence:
Ok, I'm embarrassed to say that I did in fact totally forget about this PR.  It's back on my TODO list now!

src/process/mod.rs
```diff
@@ -679,6 +681,24 @@ pub struct FDInfo {
...
+    /// Gets a file descriptor from a raw fd
+    pub fn from_raw_fd(pid: pid_t, raw_fd: i32) -> ProcResult<Self> {
+        Self::from_raw_fd_with_root("/proc", pid, raw_fd)
+    }
+
+    /// Gets a file descriptor from a raw fd based on a specified `/proc` path
+    pub fn from_raw_fd_with_root(root: impl AsRef<Path>, pid: pid_t, raw_fd: i32) -> ProcResult<Self> {
+        let path = root.as_ref().join(pid.to_string()).join("fd").join(raw_fd.to_string());
+        let link = wrap_io_error!(path, read_link(&path))?;
+        let md = wrap_io_error!(path, path.metadata())?;
```
eminence, line 693: This all looks good, except this part should be `path.symlink_metadata()`, so that the behavior here is consistent with what we're doing in `Process::fd()` (which uses [DirEntry::metadata()](https://doc.rust-lang.org/nightly/std/fs/struct.DirEntry.html#method.metadata) which doesn't dereference the symlink)

eminence (changes requested):

src/process/mod.rs
saruman9: Fixed by 78f3385

eminence:
Thank you very much!

PR #127, opened 2021-05-11
Change Process to use directory fd

### LINKED ISSUE #125: Process struct should carry an fd and use openat
Under Linux, procfs can suffer from TOCTTOU caused by pid reuse. This is evident if you take the `self_memory` example and have it run on external process. I'll show some output from strace to illustrate the problem:

```
openat(AT_FDCWD, "/proc/1234/statm", O_RDONLY|O_CLOEXEC) = 3
... snip ...
openat(AT_FDCWD, "/proc/1234/status", O_RDONLY|O_CLOEXEC) = 3
```

In between calls to `openat`, our process could be pre-empted, and then the pid could potentially be killed and re-used by another process before our process resumes. This would result in the call to `Process::status()` returning wrong information, when really it should return an error because the original process 1234 it was referencing is dead. To fix the problem, you'd want the syscalls to look more like this:

```
openat(AT_FDCWD, "/proc/1234", O_RDONLY|O_CLOEXEC) = 3
openat(3, "statm", O_RDONLY|O_CLOEXEC) = 4
openat(3, "status", O_RDONLY|O_CLOEXEC) = 4
```

The easiest way to do this would be to replace every area where file paths are joined with an intentional call to [openat](https://docs.rs/nix/0.20.0/nix/fcntl/fn.openat.html). The net effect of this would be that `Process` would no longer store `root`, it would store a handle to the directory in procfs. This should also be done when reading from entries in `/proc/[pid]/task/[tid]`. Another change that should probably happen along with this is that `process::all_processes()` should return an Iterator instead of a `Vec`, that way it will prevent a huge number of file descriptors being opened at once. This is an API break so I'm proposing it here before submitting any pull request.

zxzax:
All data accessors now use openat and readlinkat when reading procfs entries, to protect against PID reuse. Also, iterators are now used instead of `Vec` in any place where a file descriptor is held open. This is an API break, see examples/netstat.rs and examples/process_hierarchy.rs for the most obvious effects on library consumers. I also removed the `stat` field as suggested in #69. I haven't made any documentation changes yet, but I can if the approach here looks good.

A downside of this is that simple cases using `stat` are now slower. If all you need during iteration is the information from `/proc/<pid>/stat`, that now takes two additional syscalls to open and close the directory fd. We could speed that up by adding another function to iterate over the pids, and then adding more constructors to the relevant structures (`Stat`, `MemoryMap`, `Status`...) that directly reads from `/proc/<pid>/<name>`. It would have to be noted in the documentation that it's only safe to use one of those constructors while iterating, because of PID recycling; for cases where two or more files need to be read from within an individual `/proc/<pid>` directory, the fd-holding iterator from `all_processes()` must be used.

If this all sounds good, I'd like to use this as a base to add pidfd support as well, to support race-free waiting and signaling on procfs processes.

Fixes #125

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ byteorder = {version="1", features=["i128"]}
 hex = "0.4"
 flate2 = "1"
 backtrace = { version = "0.3", optional = true }
+nix = "0.20"
 
 [dev-dependencies]
 procinfo = "0.4.2"
--- a/examples/dump.rs
+++ b/examples/dump.rs
@@ -11,6 +11,7 @@ fn main() {
     };
     println!("{:#?}", prc);
 
-    println!("State: {:?}", prc.stat.state());
-    println!("RSS:   {} bytes", prc.stat.rss_bytes());
+    let stat = prc.stat().unwrap();
+    println!("State: {:?}", stat.state());
+    println!("RSS:   {} bytes", stat.rss_bytes());
 }
--- a/examples/lslocks.rs
+++ b/examples/lslocks.rs
@@ -33,7 +33,8 @@ fn main() {
         let mut found = false;
         if let Some(pid) = lock.pid {
             if let Ok(fds) = Process::new(pid).and_then(|p| p.fd()) {
-                for fd in fds {
+                for f in fds {
+                    let fd = f.unwrap();
                     if let FDTarget::Path(p) = fd.target {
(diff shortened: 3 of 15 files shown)


eminence:
I'm sorry I haven't gotten a chance to look at this yet.  I'm going to try to take a look within a week.

zxzax:
No rush. :smiley: This is still a very rough draft.

sunfishcode:
This patch was incorporated into #171, which is now merged.

PR #128, opened 2021-05-19
Derive Clone for all public structs and enums.

xfbs:
Hey, really awesome work on this!

I'm using it for some tooling and I noticed that some of the structs cannot be cloned nor copied, so I just went ahead and implemented Clone for all the types that are public. 

I don't think there is any reason not to allow the structs to be cloned.
I did not derive Copy for any of them on purpose -- that allows
non-public fields to be added later that are not Copyable.

### DIFF
--- a/src/cgroups.rs
+++ b/src/cgroups.rs
@@ -2,7 +2,7 @@ use crate::ProcResult;
 
 use super::process::Process;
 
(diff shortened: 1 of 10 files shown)


eminence:
LGTM, Thanks!

Issue #130, opened 2021-05-24
inode size is not correct on x86_64


rc-grey:
Currently inodes are defined as [u32](https://github.com/eminence/procfs/blob/master/src/process/mod.rs#L609). While it may not be likely to occur on most systems the kernel will print an unsigned long format string which can support up to eight bytes of length - right now I believe this will cause the library to ~~panic~~ return an internal error.

Relevant links:
1. https://elixir.bootlin.com/linux/latest/source/include/uapi/asm-generic/int-l64.h#L30
2. https://elixir.bootlin.com/linux/latest/source/net/socket.c#L307

rc-grey:
Will not cause library to panic - but return an internal error - thanks @eminence for clarifying.

eminence:
The `i_ino` field is defined to be an "unsigned long":

https://elixir.bootlin.com/linux/latest/source/include/linux/fs.h#L633

In C, this is defined an an unsigned 32-bit number.  Similarly, the `%lu` printf specifier also means "unsigned long".  So I think that using a `u32` is OK for storing an inode.

eminence:
Ah ha.  but hmmm:

`ULONG_MAX` on my system is indeed 8 bytes....

eminence:
Ok yeah, let's change this into a `u64` 👍

rc-grey:
@eminence Should be ready to review - in #132

PR #132, opened 2021-05-31
Changed inodes to u64, fixed a test, and updated a doc string

rc-grey:
Updated inode type, bumped version as this is a breaking change for consumers.

This addresses #130


eminence (approved):

eminence:
LGTM, thanks!

PR #135, opened 2021-07-09
Add support for /proc/<pid>/ns

Furisto:


src/process/namespaces.rs
```diff
@@ -0,0 +1,63 @@
...
+///
+/// See also the [Process::namespaces()] method
+#[derive(Debug, Clone)]
+pub struct Namespace {
+    /// Namespace type
+    ns_type: OsString,
+    /// Handle to the namespace
+    path: PathBuf,
+    /// Namespace identifier (inode number)
+    identifier: u64,
```
eminence, line 50: These 3 fields should be `pub`, right?

eminence (changes requested):
Thanks for the PR!

The namespaces manual page says this:

> If two processes are in the same namespace, then the device IDs and inode numbers of their /proc/[pid]/ns/xxx symbolic links will be the same; an application can check this using the stat.st_dev and stat.st_ino fields returned by stat(2).  

Do you think this means we should store both `st_dev` and `st_ino` in the `Namespace` struct?

src/process/namespaces.rs
Furisto: ...yes, of course! Fixed it.

Furisto:
@eminence 
You bring up an interesting point. I had only seen the inode number used to compare namespaces and tools like [lsns](https://github.com/karelzak/util-linux/blob/c2ca2837a27a3f1344904037ea691bfdbb288f02/sys-utils/lsns.c#L104-L118) do not even offer an option to display the device id for a namespace. It [seems](https://lore.kernel.org/lkml/87poky5ca9.fsf@xmission.com/) that the device id is currently not used in this capacity but _could_ be used that way in the future. I am happy to do it either way.

> Like all fstat inode comparisons to be fully accurate you need to
> compare both the st_ino and st_dev.  I reserve the right for st_dev to
> be significant when comparing namespaces.  Otherwise I might have to
> create a namespace of namespaces someday and that is ugly.

Furisto:
Added the device id and also made namespaces comparable.

Furisto:
@eminence PTAL

src/process/namespaces.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+impl PartialEq for Namespace {
+    fn eq(&self, other: &Self) -> bool {
+        // see https://lore.kernel.org/lkml/87poky5ca9.fsf@xmission.com/
+        self.identifier == other.identifier && self.device_id == other.device_id
+    }
+}
+
+impl Eq for Namespace {}
```
eminence, line 63: I think this can replaced by adding `Eq` to the list of automatic derives?  

A super minor thing, though

eminence (approved):
LGTM, thanks!

PR #137, opened 2021-07-28
Use u16 for kernel patch version number

### LINKED ISSUE #136: kernel version parsing panic
a kernel version name like `4.9.266-0.1.ac.225.84.332.metal1.x86_64` causes `procfs` to panic here

https://github.com/eminence/procfs/blob/master/src/lib.rs#L303

https://github.com/eminence/procfs/blob/d8f16ddc97413ab2fb38d23149be188ad0b750a7/src/sys/kernel/mod.rs#L66

one fix is to default to 0 values, another would be to parse the first readable number

zz85:
- Added a previously failing test
- Fixes #136

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1125,6 +1125,11 @@ mod tests {
         assert_eq!(k.major, 4);
         assert_eq!(k.minor, 9);
         assert_eq!(k.patch, 16);
+
+        let k = KernelVersion::from_str("4.9.266-0.1.ac.225.84.332.metal1.x86_64").unwrap();
+        assert_eq!(k.major, 4);
+        assert_eq!(k.minor, 9);
+        assert_eq!(k.patch, 266);
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


eminence:
Thanks, this fix looks fine to me.  But what do you think about only changing `patch: u16`, instead of changing all 3 fields to `u32`?

zz85:
i personally prefer using `u32`s since i think it isn't much of a space trade off, and that u32s is referenced in the kernel code https://github.com/torvalds/linux/blob/5bfc75d92efd494db37f5c4c173d3639d4772966/tools/testing/selftests/bpf/prog_tests/core_extern.c#L12

but i'll make a change to unblock envs this crate is crashing on

eminence:
True, but then it looks like the `KERNEL_VERSION` macro expects each field to be only u8, as it then packs everything into a single u32:

    #define KERNEL_VERSION(a,b,c) (((a) << 16) + ((b) << 8) + (c))

But we can always revisit this later.  Thanks for making this change!

PR #139, opened 2021-07-31
Adds `Process::oom_score_adj`

vrmiguel:
Hey there!

This PR adds `oom_score_adj` to `Process`. 
I don't know the minimum Linux version that has this property, though.

If merged, I plan to use this function during the testing of a small OOM killer of my own, [bustd](https://github.com/vrmiguel/bustd/) :D

### DIFF
--- a/src/process/mod.rs
+++ b/src/process/mod.rs
@@ -1109,6 +1109,22 @@ impl Process {
         Ok(from_str!(u32, oom.trim()))
     }
 
+    /// Used to adjust the badness heuristic used to select which
+    /// process gets killed in out of memory conditions.
+    /// The badness heuristic assigns a value to each candidate task ranging from 0
+    /// (never kill) to 1000 (always kill) to determine which process is targeted.  The
+    /// units are roughly a proportion along that range of allowed memory the process
+    /// may allocate from based on an estimation of its current memory and swap use.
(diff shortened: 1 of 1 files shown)


eminence (changes requested):
Hi! 👋 

Sorry it took so long to review this PR, it totally fell off my radar until now.  

I'm reading the documentation on `oom_score_adj`, and it's a little confusing.  But it says this:

> The value of `/proc/<pid>/oom_score_adj` is added to the badness score before it
> is used to determine which task to kill.  Acceptable values range from -1000
>  (`OOM_SCORE_ADJ_MIN`) to +1000 (`OOM_SCORE_ADJ_MAX`). 

So the way I read this is:  while the final OOM "score" will be somewhere between 0 and 1000, the value reported in `oom_score_adj` can be between -1000 and 1000.   So I think the `oom_score_adj` function should return a `i16` not `u32`.

Also, you could please tweak the doc string slightly:

* Put a new line after a short description of the function (then can come the full function description)
* Add "(since Linux 2.6.36)" at the very end of the doc string

Thanks!

eminence:
Hi @vrmiguel

Just checking in to see if you still if you're able to make the requested change.  If not, that's no problem at all, I can make it myself.  Thanks!

PR #140, opened 2021-08-11
Allow kernel patch version to be > 255

fabricedesre:
While it's unlikely that the major and minor version won't fit in an `u8`, I had the case of a kernel patch version overflow. Since the lazy static `KERNEL` just calls `KernelVersion::current().unwrap()` this is causing crashes.

This patch switches to a `u16` and adds a matching test.

### DIFF
--- a/src/sys/kernel/mod.rs
+++ b/src/sys/kernel/mod.rs
@@ -18,11 +18,11 @@ pub mod random;
 pub struct Version {
     pub major: u8,
     pub minor: u8,
-    pub patch: u8,
+    pub patch: u16,
 }
 
 impl Version {
-    pub fn new(major: u8, minor: u8, patch: u8) -> Version {
+    pub fn new(major: u8, minor: u8, patch: u16) -> Version {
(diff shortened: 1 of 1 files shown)


eminence:
Thanks for the patch.  The issue you identified must be fixed, though this is essentially a duplicate of #137 from @zz85.

I like your version slightly better, since it only changes the patch field to u16.  But since @zz85 got here first, they get dibs :)

Hey @zz85 -- can you take a look at my [comment](https://github.com/eminence/procfs/pull/137#issuecomment-888743828) in #137 ?  If you can switch the `patch` field to be a `u16`, that would be great.  If you can do this in the next 1 or 2 days, I'll merge your PR.   

(Since someone else ran into this problem, this priority of this issue has risen for me, so I'd like to cut a new release pretty soon)

Thank you both!

zz85:
thanks, let me try  to address #137 so a fix can  be prioritized

eminence:
Closed via #137

PR #142, opened 2021-08-24
fix compilation for 32-bit builds. It is tested for target armv7-unkn…

anstylian:
fix compilation for 32-bit builds. It is tested for target armv7-unknown-linux-gnueabihf


eminence (approved):

eminence:
Thanks!  This looks like the right fix to me.  I'll get this published as a new patch release in 1 or 2 days.

Issue #141, opened 2021-08-16
Does not work when running as 32-bit process on 64-bit machines


h33p:
When running `procfs::process::all_processes()` parsing fails in `stat.rs`, because fields overflow `usize`. Here's an example error:

```
.cargo/registry/src/github.com-1ecc6299db9ec823/procfs-0.9.1/src/process/stat.rs:306 (please report this procfs bug)
```

Changing `start_data` to be a `u64` makes the problem go away for the field, but then it fails to parse the next one:

```
.cargo/registry/src/github.com-1ecc6299db9ec823/procfs-0.9.1/src/process/stat.rs:307 (please report this procfs bug)
```

While it is not a fairly common application nowadays, a user could still run into it when downloading the wrong build of the program. I don't know what your view is on this, but this could be solved in at least a couple of ways:

1. Just replace all parsable occurances of usize to u64 (there are 60 instances of `usize` in total, 26 excluding `diskstats` module).
2. Do 1, but selectively. Ignore certain fields that would never be over 4G.
3. Create a typedef that can switch between usize and u64 as a feature. Given the way rust compiles crates, it could prove rather fragile when 2 crates expect different features, but flexible at the same time.

If this change is not worth it I'm still interested whether diskstats should be using usize, because nowadays disks are in terabytes of size, and after long runtime the stats could overflow usize on pure 32-bit machines. However, I'm not sure how Linux does it, and I could be wrong, I'm just assuming the numbers are not limited to native word size on 32-bit machines.

Either way, if it interests you I would be willing to contribute this change, just let me know which direction to go for.

eminence:
Hi, thanks for opening this issue.

I've done some testing in the past on 32-bit systems, I have to admit I've never tested a 32-bit build of procfs running on a 64-bit system.  

I also have to admit, I'm not 100% sure what to do here.

If I'm reading [this](https://github.com/torvalds/linux/blob/e22ce8eb631bdc47a4a4ea7ecf4e4ba499db4f93/include/linux/part_stat.h) correctly, the diskstat fields are `unsigned long`, which are generally 32-bits on 32-bit systems I think.  This doesn't mean that the diskstats has to follow, though.  Using `u64` here seems pretty reasonable.

[...]

PR #147, opened 2021-09-19
Switch to u64 from usize in many places

h33p:
Apologies for such a long delay, work got in the way.

I've implemented the base changes needed to fix https://github.com/eminence/procfs/issues/141. Both diskstats and stat were changed to use `u64`. I've noticed my NAS having diskstat numbers in the tens of billions so I thought the change there was appropriate. Some places I'm still not sure what to do about:

`sys/vm` `admin_reserve_kbytes` was kept as usize, as it would not make much sense to reserve billions of kilobytes.
`sys/fs` `file_max` was kept as usize, since I am not sure where would billions of files be opened.

However, both of these are rather arbitrary, and could be changed to u64. If you'd like that I'll be glad to make the change.

In the end, procfs now works just fine as a 32-bit executable on 64-bit machines.

### DIFF
--- a/src/diskstats.rs
+++ b/src/diskstats.rs
@@ -24,65 +24,65 @@ pub struct DiskStat {
     /// Reads completed successfully
     ///
     /// This is the total number of reads completed successfully
-    pub reads: usize,
+    pub reads: u64,
 
     /// Reads merged
     ///
     /// The number of adjacent reads that have been merged for efficiency.
-    pub merged: usize,
+    pub merged: u64,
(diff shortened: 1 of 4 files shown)


eminence:
Thanks for this PR, these changes look good to me.  I'm OK keeping `admin_reserve_kbytes` and `file_max` as they are.  If anyone does run into that 32-bit limit, hopefully they'll open an issue :)

eminence (approved):

PR #150, opened 2021-10-03
Remove unwraps from our lazy statics

### LINKED ISSUE #138: our lazy statics are a panic hazard
We have several lazy statics that use `.unwrap()`, thus they can panic.  This is poor API design and should be fixed.  See #136 for an example of this.

```rust
lazy_static! {
    /// The number of clock ticks per second.
    ///
    /// This is calculated from `sysconf(_SC_CLK_TCK)`.
    static ref TICKS_PER_SECOND: i64 = {
        ticks_per_second().unwrap()
    };
    /// The version of the currently running kernel.
    ///
    /// This is a lazily constructed static.  You can also get this information via
    /// [KernelVersion::new()].
    static ref KERNEL: KernelVersion = {
        KernelVersion::current().unwrap()
    };
    /// Memory page size, in bytes.
    ///
    /// This is calculated from `sysconf(_SC_PAGESIZE)`.
    static ref PAGESIZE: i64 = {
        page_size().unwrap()
    };
}
```

eminence:
Resolves #138

### DIFF
--- a/examples/dump.rs
+++ b/examples/dump.rs
@@ -12,5 +12,5 @@ fn main() {
     println!("{:#?}", prc);
 
     println!("State: {:?}", prc.stat.state());
-    println!("RSS:   {} bytes", prc.stat.rss_bytes());
+    println!("RSS:   {} bytes", prc.stat.rss_bytes().unwrap());
 }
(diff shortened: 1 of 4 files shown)


eminence:
I think this makes sense...

PR #152, opened 2021-10-14
Handle process cgroup paths containng ':'

rosds:
Cgroup paths containing ':' will cause `Process::cgroups` to return an
incomplete pathname. The reason is that parsing the fields in
`/proc/self/cgroup` is perform with `.split(':')`. If something as
follows is set:

```bash
mkdir /sys/fs/cgroup/memory/foo:bar
echo $$ > /sys/fs/cgroup/memory/foo:bar/tasks
cat /proc/self/cgroup
...
8:memory:/foo:bar
...
```

the resulting `ProcessCgroup` will have a `pathname` equal to just `/foo`.

This patch addresses this by taking the remainer of the line after the
second ':' as the pathname. This is safe because `cgroups(7)` describes
`/proc/self/cgroup` as only having 3 fields.

### DIFF
--- a/src/cgroups.rs
+++ b/src/cgroups.rs
@@ -102,7 +102,7 @@ impl Process {
                 continue;
             }
 
-            let mut s = line.split(':');
+            let mut s = line.splitn(3, ':');
             let hierarchy = from_str!(u32, expect!(s.next(), "hierarchy"));
             let controllers = expect!(s.next(), "controllers")
                 .split(',')


alfonsoros88:
@eminence thanks a lot! that was fast :)

eminence:
There were several other pending fixes that needed to be released, so your fix was just some good extra motivation to cut the release :)

Issue #156, opened 2021-11-10
Make flate2 as a optional dependencies/feature, current is only used to read /proc/config.gz


pymongo:
In many Rust projects which using procfs, mostly not use kernel_config() API

e.g. tikv project not use procfs::kernel_config()

I hope procfs's default feature is:

```
[features]
default = ["chrono", "kernel_config"]
```

let user to determine whether enable flate2 dependency

Can I write a PR for it?

eminence:
Hi,

This is a good idea.  Actually I added this feature last month, but forgot to push and make a PR.  I just opened the the PR.  Can you take a look to see if it looks OK to you?

pymongo:
resolve in #157

PR #157, opened 2021-11-10
Make the flate2 dependency optional

eminence:
CC #156

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,6 +41,7 @@
 //! The following cargo features are available:
 //!
 //! * `chrono` -- Default.  Optional.  This feature enables a few methods that return values as `DateTime` objects.
+//! * `flate2` -- Default.  Optional.  This feature enables parsing gzip compressed `/proc/config.gz` file via the `procfs::kernel_config` method.
 //! * `backtrace` -- Optional.  This feature lets you get a stack trace whenever an `InternalError` is raised.
 //!
 //! # Examples
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -671,13 +671,24 @@ pub enum ConfigSetting {
 ///
 /// If CONFIG_KCONFIG_PROC is available, the config is read from `/proc/config.gz`.
 /// Else look in `/boot/config-$(uname -r)` or `/boot/config` (in that order).
+///
+/// # Notes
+/// Reading the compress `/proc/config.gz` is only supported if the `flate2` feature is enabled
+/// (which it is by default).
+#[cfg_attr(feature = "flate2", doc = "The flate2 feature is currently enabled")]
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -665,13 +665,24 @@ pub enum ConfigSetting {
...
+#[cfg_attr(feature = "flate2", doc = "The flate2 feature is currently enabled")]
+#[cfg_attr(not(feature = "flate2"), doc = "The flate2 feature is NOT currently enabled")]
 pub fn kernel_config() -> ProcResult<HashMap<String, ConfigSetting>> {
-    use flate2::read::GzDecoder;
-
-    let reader: Box<dyn BufRead> = if Path::new(PROC_CONFIG_GZ).exists() {
-        let file = FileWrapper::open(PROC_CONFIG_GZ)?;
-        let decoder = GzDecoder::new(file);
-        Box::new(BufReader::new(decoder))
+    let reader: Box<dyn BufRead> = if Path::new(PROC_CONFIG_GZ).exists() && cfg!(feature = "flate2") {
```
pymongo, line 675: I think `&& cfg!(feature = "flate2")` is useless

pymongo: if cfg!(feature = "flate2") == false,  block would not execute

if cfg!(feature = "flate2") == true, block would execute `flate2::read::GzDecoder`

so the cfg(not(...flate2 block may never execute

pymongo: if cfg!(feature = "flate2") == false,  block would not execute

if cfg!(feature = "flate2") == true, block would execute `flate2::read::GzDecoder`

so the cfg(not(...flate2 block may never execute

pymongo:
Other prosible conditional compile logic:

1. if `feature = "flate2"` disable, not compile the pub fn kernel_config function, and all unittest relative it add cfg(feature="flate2") attr
2. like procfs::boot_times(), in chrono feature return Datetime while not chrono feature return i64, I think kernel_config not flate2 feature can return bytes, and flate2 feature return HashMap

src/lib.rs
eminence: Yes, the `#[cfg(not(feature = "flate2"))]` block is unreachable, but is needed to typecheck

pymongo:
This PR looks good to me, I can remove the flate2 dependency and speed up my project compile

eminence:
This is a breaking change, since anyone using procfs with `default-features = false` will see an incompatible change in behavior for the `kernel_config` function.  This is unfortunate, but I'm going to live with it.  We'll put a big notice in the release notes about this breaking change.  I'll also looked many crates that depend on procfs and use `default-features = false`, and none use the `kernel_config` function.

PR #161, opened 2021-11-15
Document support for /proc/pressure/{cpu,io,memory}

KernelErr:
Seems that reading PSI from /proc/pressure/* has been implemented. But it's missing in the support document.


eminence:
Good catch, thanks!

PR #162, opened 2021-11-15
Add support for /proc/net/route

pymongo:
output example:

```
[w@ww ~]$ cat /proc/net/route 
Iface   Destination     Gateway         Flags   RefCnt  Use     Metric  Mask            MTU     Window  IRTT                                          
wlp4s0  00000000        0112A8C0        0003    0       0       600     00000000        0       0       0                                             
```

```
RouteEntry { iface: "wlp4s0", destination: 0.0.0.0, gateway: 192.168.18.1, flags: 3, refcnt: 0, use: 0, metrics: 600, mask: 0.0.0.0, mtu: 0, window: 0, irtt: 0 }
```

### DIFF
--- a/src/net.rs
+++ b/src/net.rs
@@ -628,6 +628,74 @@ pub fn dev_status() -> ProcResult<HashMap<String, DeviceStatus>> {
     Ok(map)
 }
 
+/// An entry in the ipv4 route table
+#[derive(Debug, Clone)]
+pub struct RouteEntry {
+    /// Interface to which packets for this route will be sent
+    pub iface: String,
+    /// The destination network or destination host
+    pub destination: Ipv4Addr,
+    pub gateway: Ipv4Addr,
+    pub flags: u16,
+    /// Number of references to this route
+    pub refcnt: u16,
+    /// Count of lookups for the route
+    pub in_use: u16,
+    /// The 'distance' to the target (usually counted in hops)
+    pub metrics: u32,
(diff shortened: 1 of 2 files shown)


src/net.rs
```diff
@@ -628,6 +628,74 @@ pub fn dev_status() -> ProcResult<HashMap<String, DeviceStatus>> {
...
+    /// Interface to which packets for this route will be sent
+    pub iface: String,
+    /// The destination network or destination host
+    pub destination: Ipv4Addr,
+    pub gateway: Ipv4Addr,
+    pub flags: u16,
+    /// Number of references to this route
+    pub refcnt: u16,
+    /// Count of lookups for the route
+    pub r#use: u16,
```
eminence, line 643: Can you rename this to something like `count` or `in_use`?  I think it would be annoying to consumers of this field to have to use a raw identifier for this field.

eminence (changes requested):
Looks good, except for 1 minor thing.  Thanks!

src/net.rs
pymongo: Ok, I rename it to `use_` ident

eminence: Sorry, but unfortunately I don't really like `use_` either, since trailing underscores aren't part of normal rust style

pymongo: I rename it to `in_use`

eminence (approved):
Thanks!

PR #163, opened 2021-11-15
Add support for /proc/sys/kernel/{ostype,version}

KernelErr:
Add two structs `KernelType` and `KernelBuildInfo`. The work for implementing `/proc/version` should be easy as it's the combination of three substrings:

```
/proc/version
              This string identifies the kernel version that is
              currently running.  It includes the contents of
              /proc/sys/kernel/ostype, /proc/sys/kernel/osrelease, and
              /proc/sys/kernel/version.  For example:

                  Linux version 1.0.9 (quinlan@phaze) #1 Sat May 14 01:51:54 EDT 1994
```


KernelErr:
Typically `version` should be a number, but in CI, its value is `#21~20.04.1-Ubuntu`. I will change the version's type to String later.

eminence:
I took a quick skim, and it looks good so far, but I'll review this in more depth later this week.  Thanks!

eminence:
On my debian 11 machine, the value of `/proc/sys/kernel/version` is:

    1 SMP Debian 5.10.46-4 (2021-08-03)

Which the current date-parsing code doesn't really like:

    BuildInfo { version: "1", flags: {"SMP"}, time: "Debian 5.10.46-4 (2021-08-03)" }

Anything we can do about this?

KernelErr:
I created a Debian VM, and got the same result:
```
BuildInfo { version: "1", flags: {"SMP"}, time: "Debian 5.10.40-1 (2021-05-28)" }
```
Seems that `TIMESTAMP` is specified as a custom string instead of the `date` result.
```
if [ -z "$KBUILD_BUILD_TIMESTAMP" ]; then
	TIMESTAMP=`date`
else
	TIMESTAMP=$KBUILD_BUILD_TIMESTAMP
fi
```
I think we may use regex or other methods to handle this. However, I am afraid that we can not cover all formats...

eminence:
I looked around (and asked some friends), and it seems that Ubuntu, CentOS, and RHEL all use the date format that you've coded against.  Arch linux uses something similar, but not exact:

    #1 SMP PREEMPT Fri, 12 Nov 2021 19:22:10 +0000

and Debian uses this, as we discussed above:

    #1 SMP Debian 5.10.46-4 (2021-08-03)

I would prefer not to bring regex into the picture, because that's a big dependency for such a small thing.  But maybe we can try a few other date formats if the first format doesn't match?

Another option would be to rename the `time` field to something like `extra`, with a documentation note that says something like:

> This field contains any extra data from the `proc/sys/kernel/version` file.  It generally contains the build date of the kernel, but the format of the date can vary.    

We would then not even attempt to parse this date (though we could give examples in the documentation)

KernelErr:
I renamed the field `time` to `extra`. And provided a method named `extra_date` to parse date format for Ubuntu and Arch. Parsing the string from Debian would return an error like:
```
`Err` value: Other("Failed to parse extra field to date")'
```

eminence (approved):

eminence:
Thanks!

PR #166, opened 2022-02-03
Add support for `/proc/<pid>/pagemap`

EliaGeretto:
This PR adds support for accessing the page table information in `/proc/<pid>/pagemap`.

### DIFF
--- a/src/process/pagemap.rs
+++ b/src/process/pagemap.rs
@@ -13,7 +13,7 @@ const fn genmask(high: usize, low: usize) -> u64 {
 }
 
 // source: include/linux/swap.h
-pub const MAX_SWAPFILES_SHIFT: usize = 5;
+const MAX_SWAPFILES_SHIFT: usize = 5;
 
 // source: fs/proc/task_mmu.c
 bitflags! {
@@ -28,6 +28,16 @@ bitflags! {
     }
 }
 
+impl SwapPageFlags {
+    pub fn get_swap_type(&self) -> u64 {
+        (*self & Self::SWAP_TYPE).bits()
+    }
+
+    pub fn get_swap_offset(&self) -> u64 {
+        (*self & Self::SWAP_OFFSET).bits() >> MAX_SWAPFILES_SHIFT
+    }
+}
+
 bitflags! {
     pub struct MemoryPageFlags: u64 {
         const PFN = genmask(54, 0);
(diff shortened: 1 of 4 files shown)


eminence:
Hello!  Sorry for taking so long to review this.  Quick observation to start, there's a build problem in the tests (also, no clue why github isn't running the CI tests...)

```
error[E0277]: the trait bound `std::ops::Range<u64>: std::ops::RangeBounds<usize>` is not satisfied
   --> src/process/tests.rs:269:45
    |
269 |     let page_infos = pagemap.get_range_info(start_page..end_page).unwrap();
    |                              -------------- ^^^^^^^^^^^^^^^^^^^^ the trait `std::ops::RangeBounds<usize>` is not implemented for `std::ops::Range<u64>`
    |                              |
    |                              required by a bound introduced by this call
    |
```

EliaGeretto:
@eminence I have fixed the merge conflict rebasing on `master`.

eminence:
@EliaGeretto thanks!  I'll take a look this weekend!

src/process/pagemap.rs
```diff
@@ -0,0 +1,185 @@
...
+const MAX_SWAPFILES_SHIFT: usize = 5;
+
+// source: fs/proc/task_mmu.c
+bitflags! {
+    /// Represents the fields and flags in a page table entry for a swapped page.
+    pub struct SwapPageFlags: u64 {
+        /// Swap type if swapped
+        const SWAP_TYPE = genmask(MAX_SWAPFILES_SHIFT - 1, 0);
+        /// Swap offset if swapped
+        const SWAP_OFFSET = genmask(54, MAX_SWAPFILES_SHIFT);
```
eminence, line 25: If I'm understanding this correctly, these two entries are bitmasks, and the only reason they are declared like this is so that the call to `from_bits_truncate` will not discard these bits.  Is that right?

If so, then I'm wondering if we can improve the doc-comment on these to indicate that a user would never need to test to see if `SWAP_TYPE` or `SWAP_OFFSET` are set (but instead should use `get_swap_type()` and `get_swap_offset()`.

I guess another option might be to use `from_bits_unchecked` and store these masks outside of the bitflags (though I'd be worried about certain bitflag options later losing these bits).

Can these marks be marked as `#[doc(hidden)]` to reduce confusion for the user?

EliaGeretto:
I have added `#[doc(hidden)]` to `SWAP_TYPE` and `SWAP_OFFSET`, but also `PFN` and `SWAP`. In the case of `PFN` the same reasoning applies. In the case of `SWAP`, the information in that bit is already encoded in the type of the bitmask (`SwapPageFlags` vs `MemoryPageFlags`), it should never be needed.

src/process/pagemap.rs
EliaGeretto: See comment below.

eminence:
Makes sense.  Thanks!

PR #169, opened 2022-03-14
Fix a race condition in the `test_task_runsinglethread` test.

sunfishcode:
`test_task_runsinglethread` spawns a thread which intends to do no I/O,
and checks that its `rchar` io value is 0. However, glibc sometimes
reads from /sys/devices/system/cpu/online, which sometimes happens on
this thread, and that causes a non-zero `rchar` value. The fix here is
to only require `rchar` to be less than the number of bytes that the
other thread reads. Also, check that `read_bytes` returns 0 since it
doesn't include bytes read from sysfs, and check that `wchar` is 0.


eminence:
Nice fix, thank you!

PR #171, opened 2022-03-15
Revive RFC: Change Process to use directory fd

### LINKED ISSUE #125: Process struct should carry an fd and use openat

sunfishcode:
This updates and polishes the patch from #127, which seems to be abandoned, as the author is listed as @ghost. It uses `openat` and `readlinkat` to protect against PID reuse.

This also makes two additional changes: it adds a dependency on [rustix](https://crates.io/crates/rustix), which provides safe APIs for `openat`, `readlinkat`, and so on, and it bumps the MSRV to 1.48. I know you just recently bumped the MSRV to 1.42, but 1.48 is worth considering, as it's the version of Rust currently in Debian stable, it's newer than the version in the oldest still-supported Fedora release, and it eliminates the need for special instructions for the hex and bitflags dependencies. That said, people have different needs, and of course it's your call as to whether these changes are appropriate for procfs.

Fixes #125.

### DIFF
--- a/src/net.rs
+++ b/src/net.rs
@@ -20,8 +20,7 @@
 //! # use std::collections::HashMap;
 //! let all_procs = procfs::process::all_processes().unwrap();
 //!
-//! // build up a map between socket inodes and processes:
-//! // build up a map between socket inodes and processes:
+//! // build up a map between socket inodes and process stat info:
 //! let mut map: HashMap<u64, Stat> = HashMap::new();
 //! for p in all_procs {
 //!     let process = p.unwrap();
--- a/src/net.rs
+++ b/src/net.rs
@@ -16,17 +16,19 @@
 //! > cargo run --example=netstat
 //!
 //! ```rust
-//! # use procfs::process::{Process, FDTarget};
+//! # use procfs::process::{FDTarget, Stat};
 //! # use std::collections::HashMap;
 //! let all_procs = procfs::process::all_processes().unwrap();
 //!
 //! // build up a map between socket inodes and processes:
-//! let mut map: HashMap<u64, &Process> = HashMap::new();
-//! for process in &all_procs {
-//!     if let Ok(fds) = process.fd() {
+//! // build up a map between socket inodes and processes:
+//! let mut map: HashMap<u64, Stat> = HashMap::new();
+//! for p in all_procs {
+//!     let process = p.unwrap();
+//!     if let (Ok(stat), Ok(fds)) = (process.stat(), process.fd()) {
 //!         for fd in fds {
-//!             if let FDTarget::Socket(inode) = fd.target {
-//!                 map.insert(inode, process);
+//!             if let FDTarget::Socket(inode) = fd.unwrap().target {
+//!                 map.insert(inode, stat.clone());
 //!             }
 //!         }
 //!     }
@@ -41,8 +43,8 @@
 //!     let local_address = format!("{}", entry.local_address);
 //!     let remote_addr = format!("{}", entry.remote_address);
 //!     let state = format!("{:?}", entry.state);
-//!     if let Some(process) = map.get(&entry.inode) {
-//!         println!("{:<26} {:<26} {:<15} {:<12} {}/{}", local_address, remote_addr, state, entry.inode, process.stat.pid, process.stat.comm);
+//!     if let Some(stat) = map.get(&entry.inode) {
+//!         println!("{:<26} {:<26} {:<15} {:<12} {}/{}", local_address, remote_addr, state, entry.inode, stat.pid, stat.comm);
 //!     } else {
 //!         // We might not always be able to find the process associated with this socket
 //!         println!("{:<26} {:<26} {:<15} {:<12} -", local_address, remote_addr, state, entry.inode);
--- a/src/process/mod.rs
+++ b/src/process/mod.rs
@@ -698,7 +698,7 @@ impl FDInfo {
     /// Gets a file descriptor from a directory fd and a path relative to it.
     ///
     /// `base` is the path to the directory fd, and is used for error messages.
-    pub fn from_process_at<P: AsRef<Path>, Q: AsRef<Path>>(
+    fn from_process_at<P: AsRef<Path>, Q: AsRef<Path>>(
         base: P,
         dirfd: BorrowedFd,
         path: Q,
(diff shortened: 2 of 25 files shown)


eminence:
Hey, this is pretty neat, thanks for the PR.  There's a lot to unpack (and to review) here, so please bear with me...

I took a quick skim and despite the large diff, everything is pretty readable.  But I do want to spend some more time looking at all the changes.  A few brief comments for now:

* I think I'm fine bumping the MSRV to 1.48, given what you said about the versions currently shipping in Debian and Fedora (thanks for those details).  1.48 is already more than 1 year old, and if anyone really needs to use something older, I'm OK asking them to stay on `v0.12` (if merged, this PR would be part of a `v0.13` release).
* The change to use `openat` and friends looks nice.   In particular, the simplification of `Process::is_alive` is a good demonstration of the benefits of this approach, I like it.
* I'm not opposed to the port to `rustix`, but I do want to study it a bit more.  In particular, a small number of `rustix` types are exposed as part of the public API of this crate, so the port to `rustix` is not quite an entirely internal implementation details.

sunfishcode:
>     * I'm not opposed to the port to `rustix`, but I do want to study it a bit more.  In particular, a small number of `rustix` types are exposed as part of the public API of this crate, so the port to `rustix` is not quite an entirely internal implementation details.

Some of those things were APIs which don't actually need to be public, and others were things that could be avoided. I've now fixed them, so rustix doesn't appear in the public API at all.

eminence:
Thanks, I appreciate that

src/process/task.rs
```diff
@@ -154,7 +162,7 @@ mod tests {
             if stat.comm == "one" && status.name == "one" {
                 found_one = true;
                 assert!(io.rchar >= bytes_to_read);
-                assert!(stat.utime >= 50, "utime({}) too small", stat.utime);
+                assert!(stat.utime >= 1, "utime({}) too small", stat.utime);
```
eminence, line 165: This change was just to get the tests more stable, right?  (The utime accounting here is known to be a bit flakey, and I've had to fuss with these tests quite often)

src/process/mod.rs
```diff
@@ -675,34 +672,69 @@ impl FromStr for FDTarget {
...
+            root,
+            rustix::fs::openat(
+                dirfd,
+                p,
+                OFlags::NOFOLLOW | OFlags::PATH | OFlags::CLOEXEC,
+                Mode::empty()
+            )
+        )
+        .map(|fdfd| File::from_fd(fdfd.into()))?;
+        let fdfd = file.as_fd();
```
eminence, line 724: Why does this have to go through `File::from_fd()`?   Can't just used the `OwnedFd` returned by `openat`, and borrow it when calling `statat` and `readlinkat` ?

eminence (approved):

src/process/task.rs
sunfishcode: This came from #127, so I'm not sure of the original reason for it. I'll revert it.

src/net.rs
```diff
@@ -16,17 +16,19 @@
...
-//! # use procfs::process::{Process, FDTarget};
+//! # use procfs::process::{FDTarget, Stat};
 //! # use std::collections::HashMap;
 //! let all_procs = procfs::process::all_processes().unwrap();
 //!
 //! // build up a map between socket inodes and processes:
-//! let mut map: HashMap<u64, &Process> = HashMap::new();
-//! for process in &all_procs {
-//!     if let Ok(fds) = process.fd() {
+//! // build up a map between socket inodes and processes:
```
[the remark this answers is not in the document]
sunfishcode: Fixed.

src/process/mod.rs
sunfishcode: Ah, right, we don't need a `FIle` here. That makes this code much simpler.

sunfishcode:
Is there anything else you'd like to see in this PR?

eminence:
No, sorry for my delinquency here.  I did spend some time trying to update a project to using this new version, and I will say that the fact that `Process` is no longer clone and the fact that there is no longer a pre-populated `stat` field was surprisingly annoying.   I think I'm going to want to think about a new quality-of-life APIs before releasing this.

But that of course doesn't need to help up this PR.   

One final question for you:  How closely should I be following `rustix` versions?  Do you think the current minimum version of `0.34.0` will be fine for the foreseeable future?

src/process/mod.rs
```diff
@@ -1239,55 +1299,120 @@ impl Process {
...
+        )
+    )?;
+    let dir = wrap_io_error!(root, rustix::fs::Dir::from(dir))?;
+    Ok(ProcessesIter { inner: dir })
 }
 
-/// Return a list of all processes based on a specified `/proc` path
-///
-/// If a process can't be constructed for some reason, it won't be returned in the list.
-pub fn all_processes_with_root(root: impl AsRef<Path>) -> ProcResult<Vec<Process>> {
```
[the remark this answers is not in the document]
sunfishcode: Oops, most likely I lost track of it while resolving a merge conflict. I've now opened https://github.com/eminence/procfs/pull/184 to reintroduce it.

PR #173, opened 2022-06-02
Build with cargo update -Z minimal-versions

xfix:
Upgrading crates to actual minimal versions so that this crate would build with `cargo update -Z minimal-versions`.


eminence:
LGTM, thanks!

PR #176, opened 2022-06-28
Process: Add support for network

arilou:
It's required to access the process /net which might point to a network
private namespace.


eminence:
Thanks!

PR #177, opened 2022-07-05
Remove obsolete document about stat field

dalance:
`stat` pre-load seems to be removed at 0.13.0.


eminence:
Nice catch, thank you!

PR #178, opened 2022-07-05
Process: Correct AUXV to support 64bit Linux

arilou:


eminence:
Thanks.  I'm embarrassed to say that function has been broken ever since I added it 4 years ago (307aa874720e9a0d9056ba9f015b98cec2f9336c)!

arilou:
Heh no problem, thanks for merging so quickly :) 
I was hoping if you are willing to expose /proc/X/mem file out, so we can read/write to a process virtual memory
(now that AUXV is valid it makes more sense)

What do you think?

arilou:
That's awesome thank you!

PR #180, opened 2022-07-06
Add suport for /proc/[pid]/task/[tid]/children

zmjackson:
While using this library I came across the need to find the children of a process. I saw that it was not supported, but it seemed fairly easy to implement so I took a shot at it.

I'm new to Rust so I welcome any suggestions.

### DIFF
--- a/src/process/task.rs
+++ b/src/process/task.rs
@@ -77,6 +77,9 @@ impl Task {
         Schedstat::from_reader(FileWrapper::open_at(&self.root, &self.fd, "schedstat")?)
     }
 
+    /// Thread children from `/proc/<pid>/task/<tid>/children`
+    ///
+    /// This data will be unique per task.
     pub fn children(&self) -> ProcResult<Vec<u32>> {
         let mut buf = String::new();
         let mut file = FileWrapper::open_at(&self.root, &self.fd, "children")?;
--- a/src/process/task.rs
+++ b/src/process/task.rs
@@ -94,6 +94,7 @@ impl Task {
 #[cfg(test)]
 mod tests {
     use crate::process::Io;
+    use std::process;
     use std::sync::{Arc, Barrier};
 
     #[test]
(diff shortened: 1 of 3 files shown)


src/process/task.rs
```diff
@@ -188,4 +206,30 @@ mod tests {
...
+
+    #[test]
+    fn test_task_children() {
+        // Use tail -f /dev/null to create two infinite processes
+        let mut command = process::Command::new("tail");
+        command.arg("-f").arg("/dev/null");
+        let (mut child1, mut child2) = (command.spawn().unwrap(), command.spawn().unwrap());
+
+        // Cargo tests run on multiple threads under the same process by default
+        // There doesn't seem to be an easy way to get the current thread ID
```
eminence, line 218: What about https://docs.rs/rustix/latest/rustix/thread/fn.gettid.html to get the thread ID?

zmjackson: Good point. Didn't realize rustix has that utility. I'll change it.

eminence:
You can ignore the CI failure (or you can rebase to get a fix for that)

src/process/task.rs
```diff
@@ -75,11 +76,28 @@ impl Task {
     pub fn schedstat(&self) -> ProcResult<Schedstat> {
         Schedstat::from_reader(FileWrapper::open_at(&self.root, &self.fd, "schedstat")?)
     }
+
+    /// Thread children from `/proc/<pid>/task/<tid>/children`
+    ///
+    /// This data will be unique per task.
```
eminence, line 82: I'd love to see these docs expanded a bit.  In particular, the `procfs(5)` man page lists some caveats that look important and I'd like to see them replicated here.

eminence:
> While using this library I came across the need to find the children of a process. I saw that it was not supported, but it seemed fairly easy to implement so I took a shot at it.

I should also note that using `/proc/[pid]/task/[tid]/children` is not really the recommended way to get the children of a process.   Generally you have to look at each processes parent process and build up a list of children processes from that.  See also [this example](https://github.com/eminence/procfs/blob/master/examples/process_hierarchy.rs)

eminence:
Could you please rebase on top of the latest master branch (instead of merging it into your feature branch)?  This will result is a much cleaner history.  If you're not sure how to do this, I'd be happy to help explain the steps.

zmjackson:
Sorry, I don't have a lot of experience with forked projects. I tried to rebase but clearly it didn't work. What steps should I take to get it in the right state?

eminence:
Oh, I think I see what might have happened.  Correct me if I'm wrong, but I think your successfully rebased, but when you went to push your branch, git told you that it wasn't a fast-forward, and it recommended that you run `git pull`.  Unfortunately, running `git pull` after a rebase is almost always the *wrong* thing to do.  (Not really your fault, I've been annoyed at git for *years* that it continues to make this recommendation)

I hope you don't mind, but I've fixed this for you by force-pushing to your `task-children` branch.  Be sure to fetch and `git reset` your local branch before re-using it.

Since I did a force-push, can you quickly skim the commits in this PR to make sure they look OK to you?

zmjackson:
Sorry for the delay, that is exactly what happened. Thanks so much for the help! Everything looks good.

PR #185, opened 2022-07-17
Process: namespaces: Change from Vec to HashMap

arilou:
This change breaks the API for process namespaces, but it provides
a more efficient way of working with the namespaces of a process.

### DIFF
--- a/src/process/namespaces.rs
+++ b/src/process/namespaces.rs
@@ -1,5 +1,6 @@
 use rustix::fs::AtFlags;
 use std::{
+    collections::HashMap,
     ffi::OsString,
     fs::{self},
     path::PathBuf,
@@ -12,9 +13,10 @@ use super::Process;
 impl Process {
     /// Describes namespaces to which the process with the corresponding PID belongs.
     /// Doc reference: https://man7.org/linux/man-pages/man7/namespaces.7.html
-    pub fn namespaces(&self) -> ProcResult<Vec<Namespace>> {
+    /// The namespace type is the key for the HashMap, i.e 'net', 'user', etc.
+    pub fn namespaces(&self) -> ProcResult<HashMap<OsString, Namespace>> {
         let ns = self.root.join("ns");
-        let mut namespaces = Vec::new();
+        let mut namespaces = HashMap::new();
         for entry in fs::read_dir(ns)? {
             let entry = entry?;
             let path = entry.path();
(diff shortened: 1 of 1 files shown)


eminence:
LGTM, except for the small logic fix

arilou:
Cool can you please tell me what would you like me to change? Which logic?

eminence (changes requested):

eminence:
Oops, sorry about that.  I forgot to click the "submit review" button

src/process/namespaces.rs
```diff
@@ -12,22 +13,27 @@ use super::Process;
...
+            namespaces
+                .insert(
+                    ns_type,
+                    Namespace {
+                        path,
+                        identifier: stat.st_ino,
+                        device_id: stat.st_dev,
+                    },
+                )
+                .expect("NsType already in map");
```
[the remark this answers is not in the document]
arilou: Sure will fix the expect with a proper error propagation, I'm sorry I'm not sure I understand which part is backwards?
We insert to the HashMap, using the ns_type as Key, and the Value is the Namespace structure.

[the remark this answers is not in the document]
eminence: Take a look at the CI test failure.

The first time we insert something into the HashMap, it will return `None` to indicate that the key doesn't exist in the map already.  But this causes the `expect` to panic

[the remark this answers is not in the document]
arilou: Oh I see thank you sorry! missed something silly like that will fix :)

src/process/namespaces.rs
```diff
@@ -39,8 +45,6 @@ impl Process {
 /// See also the [Process::namespaces()] method
 #[derive(Debug, Clone)]
 pub struct Namespace {
-    /// Namespace type
-    pub ns_type: OsString,
```
eminence, line 43: Also, for better compatibility, I'm inclined to keep the `ns_type` field.  For people who are currently using the `Vec<_>` interface, including the `ns_type` in the struct is more useful, at the expense of a small amount of memory duplication

arilou: No problem will do

src/process/namespaces.rs
```diff
@@ -12,22 +13,36 @@ use super::Process;
...
+                )
+                .map_or_else(
+                    || Ok(()),
+                    |n| {
+                        Err(build_internal_error!(format!(
+                            "NsType appears more than once {:?}",
+                            n.ns_type
+                        )))
+                    },
+                )?;
```
eminence, line 45: Sorry to be a pain about this, but I think this can be written much more cleanly by just using a simple `if` statement:

```suggestion
            if namespaces
                .insert(
                    ns_type.clone(),
                    Namespace {
                        ns_type,
                        path,
                        identifier: stat.st_ino,
                        device_id: stat.st_dev,
                    },
                )
                .is_some()
            {
                return Err(build_internal_error!(format!("Duplicate namespace type {:?}", ns_type)));
            }
```

arilou: No problem, pushed a new revision, please let me know if there is any other change you would like to have.

eminence (approved):

eminence:
Thanks!

eminence:
This change has been published as version `v0.14.0`

Issue #188, opened 2022-07-22
procfs::process::all_processes() can blow "open files" ulimit on a system with a large number of processes


obaudys:
Looks like the changes in pull request #171 make the `procfs::process::Process` struct hold an open file descriptor for the corresponding `/proc/<pid>`  directory,  and stats are then lazily loaded with methods.

Unfortunately, keeping a lot of open file descriptors around can hit the "max open files" ulimit - it is 1024 on most linux installs by default.  This happened to me today while playing around with [nushell](https://github.com/nushell/nushell), when the `ps` command did not work, since it had hit the open files limit.

The code in question is [this snippet](https://github.com/nushell/nushell/blob/c66b97126f71e24a99a2f265b3f5b5420488277d/crates/nu-system/src/linux.rs#L83-L126), where `procfs::process::all_processes()` is called, then more `Process` instances are created from there which really explodes the number of open file descriptors.

Unfortunately I think the change in #171 - ie. wrapping an open file descriptor in a struct and holding it open until such time as it is needed - is a bad idea.  Why not just hold name of the `/prod/<pid>` directory as a String?  Convention dictates that a file is opened, data is read, and the file is closed in rapid succession.  This is effectively how it is being done by the methods in `impl Process`, _except for the open file descriptor for the directory!_

To reproduce the issue,  I offer this butchered version of `examples/ps.rs`:
```rust
#![allow(clippy::print_literal)]

use procfs::process::Process;

extern crate procfs;

fn main() {
    let mestat = procfs::process::Process::myself().unwrap().stat().unwrap();
    let tps = procfs::ticks_per_second().unwrap();

    println!("{: >10} {: <8} {: >8} {}", "PID", "TTY", "TIME", "CMD");

    let tty = format!("pty/{}", mestat.tty_nr().1);
    let mut persist = Vec::new();
    match procfs::process::all_processes() {
        Ok(all_procs) => {
            for p in all_procs {
                match p {
                    Ok(prc) => {
                        if let Ok(stat) = prc.stat() {
                            // total_time is in seconds
                            let total_time = (stat.utime + stat.stime) as f32 / (tps as f32);
                            println!("{: >10} {: <8} {: >8} {}", stat.pid, tty, total_time, stat.comm);
                        }
                        let newproc = match Process::new(prc.pid()) {
                            Ok(p) => p,
                            Err(e) => panic!("{}",e) // cannot create new Process!
                        };
                        // don't let processes go out of scope, to trigger "Too many open files"
[...]

PR #190, opened 2022-07-25
Add some docs on the Process object

eminence:
CC #188

### DIFF
--- a/src/process/mod.rs
+++ b/src/process/mod.rs
@@ -755,6 +755,20 @@ impl std::fmt::Debug for FDInfo {
 }
 
 /// Represents a process in `/proc/<pid>`.
+///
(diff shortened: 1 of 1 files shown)


eminence:
If you have time for a quick review @obaudys, I'd be grateful.

obaudys (changes requested):
Looks good to me; it neatly spells out what is going on and why.

You seem to have closed the `impl` block and then reopened it:
```
}

impl Process {
```
Is there any reason for that?  Asking as a rust newbie.

eminence:
It's subtle, but I added some docs to the impl block on line 779.  This isn't all that well known (and honestly not used that often), but you can add documentation to the impl block itself, as in:

```rust
/// This is some docs
impl Foo {
    /// Create a new thing
    pub fn new() -> Self { todo!() }
}

PR #192, opened 2022-08-01
Process: Namespace: Use openat instead of building a path

arilou:
This should resolve a case of a PID reuse

### DIFF
--- a/src/process/mod.rs
+++ b/src/process/mod.rs
@@ -1405,6 +1405,12 @@ impl Process {
         let file = FileWrapper::open_at(&self.root, &self.fd, "mem")?;
         Ok(file.inner())
     }
+
+    /// Returns a file which is part of the process proc structure
+    pub fn open_relative(&self, path: &str) -> ProcResult<File> {
+        let file = FileWrapper::open_at(&self.root, &self.fd, path)?;
+        Ok(file.inner())
+    }
 }
 
 /// The result of [`Process::fd`], iterates over all fds in a process
--- a/src/process/namespaces.rs
+++ b/src/process/namespaces.rs
@@ -1,10 +1,5 @@
-use rustix::fs::AtFlags;
-use std::{
-    collections::HashMap,
-    ffi::OsString,
-    fs::{self},
-    path::PathBuf,
-};
+use rustix::fs::{AtFlags, Mode, OFlags};
+use std::{collections::HashMap, ffi::OsString, path::PathBuf};
 
 use crate::ProcResult;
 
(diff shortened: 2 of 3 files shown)


src/process/namespaces.rs
```diff
@@ -15,13 +10,33 @@ impl Process {
...
+                Mode::empty()
+            )
+        )?;
+        let dir = wrap_io_error!(self.root.join("ns"), rustix::fs::Dir::read_from(&dir_ns))?;
+        for entry in dir {
+            let entry = entry.map_err(|_| build_internal_error!(format!("Unable to get ns dir entry")))?;
+            match entry.file_name().to_string_lossy().as_ref() {
+                "." | ".." => continue,
+                _ => {}
+            };
```
eminence, line 29: I think this might be a bit more efficient:

```suggestion
            match entry.file_name().to_bytes() {
                b"." | b".." => continue,
                _ => {}
            };
```

src/process/namespaces.rs
```diff
@@ -15,13 +10,33 @@ impl Process {
...
+
+            let path = self
+                .root
+                .join("ns")
+                .join(entry.file_name().to_string_lossy().to_string());
```
eminence, line 34:
```suggestion
                .join(entry.file_name().to_string_lossy().as_ref());
```

src/process/namespaces.rs
```diff
@@ -15,13 +10,33 @@ impl Process {
...
+            let ns_type = OsString::from(entry.file_name().to_string_lossy().to_string());
```
eminence, line 35:
```suggestion
            let ns_type = OsString::from(entry.file_name().to_string_lossy().as_ref());
```

eminence (changes requested):
LGTM, though please take a look at some of the suggestions I made.  Thanks!

src/process/namespaces.rs
```diff
@@ -15,13 +10,33 @@ impl Process {
...
+            let dir_ns = dir_ns
+                .try_clone()
+                .map_err(|_| build_internal_error!(format!("Unable to clone ns dir")))?;
+            let stat = rustix::fs::statat(dir_ns, entry.file_name(), AtFlags::empty())
```
eminence, line 39: I don't think this clone is necessary

```suggestion
            let dir_ns = dir_ns
                .try_clone()
                .map_err(|_| build_internal_error!(format!("Unable to clone ns dir")))?;
            let stat = rustix::fs::statat(dir_ns, entry.file_name(), AtFlags::empty())
```
```suggestion
            let stat = rustix::fs::statat(&dir_ns, entry.file_name(), AtFlags::empty())
```

src/process/mod.rs
```diff
@@ -771,7 +771,7 @@ impl std::fmt::Debug for FDInfo {
 ///
 #[derive(Debug)]
 pub struct Process {
-    fd: OwnedFd,
+    pub fd: OwnedFd,
```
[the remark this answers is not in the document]
arilou: Change the order of the patches and using open_relative instead

src/process/namespaces.rs
arilou: Done

src/process/namespaces.rs
arilou: Done

src/process/namespaces.rs
arilou: Done

src/process/namespaces.rs
arilou: Done

arilou:
Thank you for the review, I was thinking perhaps to add a derive Serialize/Deserialize to some of the public structs, does that sound ok to you?

eminence (approved):

eminence:
> I was thinking perhaps to add a derive Serialize/Deserialize to some of the public structs, does that sound ok to you?

That sounds OK to me, but please put them behind a off-by-default feature flag

PR #193, opened 2022-08-08
add serde serialize/deserialize derives for public types

eliad-wiz:
put them under non-default feature "serde1".

exclude some "native" types (such as OwnedFd), that can't
be serialized/deserialized properly.

Signed-off-by: Eliad Peller <eliad.peller@wiz.io>


eminence:
LGTM on a quick skim, though I'll review more carefully tonight.  I'm curious to know more about your use-case, if you don't mind sharing some details.

eliad-wiz:
we simply collect information about some processes, and would like to be able to serialize/deserialize the information easily (e.g. for consumption by other tools)

eminence:
I spent some more time looking at this, and this continues to look good to me.  Since this crate is predominately about parsing structured data, it makes sense to provide serde impls.  I can't guarantee that I'll remember to add the derives on future structs, so don't hesitate to get in touch if something is missing

PR #196, opened 2022-08-13
Adjust serde derives on InternalError struct

eminence:
The optional `backtrace` feature is slightly problematic when deriving
Serialize and Deserialize.  By default, the `backtrace` crate doesn't support
serde, but it has an optional feature to do so.

Ideally, this problem would be solved by adjusting the dependency in
Cargo.toml:

```toml
[features]
serde1 = ["serde", "backtrace?/serde"]
```

But this syntax is only available in rust 1.60, while the MSRV for
procfs is 1.48.

So the until we bump our MSRV, the `backtrace` field will be skiped for
serialization.  Since this type has a `Default` impl, we could in theory
deserialize it, but that doesn't seem to make much sense, so the
`InternalError` struct no longer derives Deserialize.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -462,12 +462,13 @@ pub enum ProcError {
 ///
 /// If you compile with the optional `backtrace` feature (disabled by default),
 /// you can gain access to a stack trace of where the error happened.
-#[cfg_attr(feature = "serde1", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde1", derive(Serialize))]
 pub struct InternalError {
     pub msg: String,
     pub file: &'static str,
     pub line: u32,
     #[cfg(feature = "backtrace")]
+    #[cfg_attr(feature = "serde1", serde(skip))]
     pub backtrace: backtrace::Backtrace,
 }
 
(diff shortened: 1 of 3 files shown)


eminence:
@arilou or @eliad-wiz -- would you mind reviewing this?

src/lib.rs
```diff
@@ -462,12 +462,13 @@ pub enum ProcError {
 ///
 /// If you compile with the optional `backtrace` feature (disabled by default),
 /// you can gain access to a stack trace of where the error happened.
-#[cfg_attr(feature = "serde1", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde1", derive(Serialize))]
```
eliad-wiz, line 465: doesn't matter much, but i guess that instead of completely removing Deserialize you can limit it to something like:
```
#[cfg_attr(feature = "serde1", derive(Serialize))]
#[cfg_attr(all(feature = "serde1", not(feature = "backtrace")), derive(Deserialize))]
```

eliad-wiz (commented):
looks good to me. thanks for the fixes!

src/lib.rs
eminence: Yeah, that could work, but I think I'm going to opt for the simplier implementation (just removing `Deserialize`) until someone has a concrete usecase where they need it

eminence:
Thanks for the review!

PR #204, opened 2022-09-18
Fix process iterator when used with a custom root 

ludo-c:
Hello,

My context is a lxc container with a limited version of `/proc` containing only processes that are running in the container.
The full proc mounted in another directory, `/proc-full`.

When I call `all_processes_with_root` with the root `/proc-full`, the iterator creates `Process` with the "default" constructor instead of the one with `with_root`: https://github.com/eminence/procfs/blob/master/src/process/mod.rs#L1541
The creation of the Process fails silently as there is no check for that and `all_processes_with_root` returns successfully, but without processes.
If I change that call with `Process::new_with_root`, it works.

As Rust is quite new to me I'm not sure that it is a good solution, can you have a look at it please ?

Thank you very much !

PS: I left some few commits from my last PR:
- In the tests I change the `failure` crate to `anyhow`. `failure` has been marked as deprecated : https://github.com/rust-lang-deprecated/failure
- add a new test for the ` loadavg_from_reader`
- Fix a `cargo clippy` warning

The test `sys::kernel::random::tests::test_poolsize` fails but it seems that is was failing before these modifications too.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1181,6 +1181,19 @@ mod tests {
     }
 
     #[test]
+    fn test_loadavg_from_reader() -> ProcResult<()> {
+        let load_average = LoadAverage::from_reader("2.63 1.00 1.42 3/4280 2496732".as_bytes())?;
+
+        assert_eq!(load_average.one, 2.63);
+        assert_eq!(load_average.five, 1.00);
+        assert_eq!(load_average.fifteen, 1.42);
+        assert_eq!(load_average.max, 4280);
+        assert_eq!(load_average.cur, 3);
+        assert_eq!(load_average.latest_pid, 2496732);
+        Ok(())
+    }
+
+    #[test]
     fn test_from_str() -> ProcResult<()> {
         assert_eq!(from_str!(u8, "12"), 12);
         assert_eq!(from_str!(u8, "A", 16), 10);
--- a/src/process/mod.rs
+++ b/src/process/mod.rs
@@ -862,7 +862,7 @@ impl Process {
     }
 
     /// Is this process still alive?
-    /// 
+    ///
     /// Processes in the Zombie or Dead state are not considered alive.
     pub fn is_alive(&self) -> bool {
         if let Ok(stat) = self.stat() {
@@ -1516,7 +1516,10 @@ pub fn all_processes_with_root(root: impl AsRef<Path>) -> ProcResult<ProcessesIt
         )
     )?;
     let dir = wrap_io_error!(root, rustix::fs::Dir::read_from(dir))?;
-    Ok(ProcessesIter { inner: dir })
+    Ok(ProcessesIter {
+        root: PathBuf::from(root),
+        inner: dir,
+    })
 }
 
 /// An iterator over all processes in the system.
@@ -1528,6 +1531,7 @@ pub fn all_processes_with_root(root: impl AsRef<Path>) -> ProcResult<ProcessesIt
 /// for more information.
 #[derive(Debug)]
 pub struct ProcessesIter {
+    root: PathBuf,
     inner: rustix::fs::Dir,
 }
 
@@ -1538,7 +1542,7 @@ impl std::iter::Iterator for ProcessesIter {
             match self.inner.next() {
                 Some(Ok(entry)) => {
                     if let Ok(pid) = i32::from_str(&entry.file_name().to_string_lossy()) {
-                        if let Ok(proc) = Process::new(pid) {
+                        if let Ok(proc) = Process::new_with_root(self.root.join(pid.to_string())) {
                             break Some(Ok(proc));
                         }
                     }


ludo-c:
I removed the commit with the documentation change.

```diff
-#![deny(broken_intra_doc_links)]
+#![deny(rustdoc::broken_intra_doc_links)]
```

but it brokes the build, the commit has been removed from the PR.

src/process/mod.rs
```diff
@@ -1528,6 +1531,7 @@ pub fn all_processes_with_root(root: impl AsRef<Path>) -> ProcResult<ProcessesIt
 /// for more information.
 #[derive(Debug)]
 pub struct ProcessesIter {
+    root: String,
```
tatref, line 1534: This should probably be `PathBuf` or `OsString`, because a path on Linux can be any bytes, whereas a String is valid UTF-8.

src/process/mod.rs
```diff
@@ -1538,8 +1542,10 @@ impl std::iter::Iterator for ProcessesIter {
             match self.inner.next() {
                 Some(Ok(entry)) => {
                     if let Ok(pid) = i32::from_str(&entry.file_name().to_string_lossy()) {
-                        if let Ok(proc) = Process::new(pid) {
-                            break Some(Ok(proc));
+                        let process_root = PathBuf::from(&self.root).join(pid.to_string());
+                        match Process::new_with_root(process_root) {
+                            Ok(proc) => break Some(Ok(proc)),
+                            Err(e) => break Some(Err(e)),
```
tatref, line 1548: As it works now, the function will loop until a valid Process is found (Err is simply ignored). If you explicitly  handle the error, you can remove the `loop`, and the `break`s are also not useful anymore.

I prefer your wait of explicitly handling the error, but this changes the behavior a bit

src/process/mod.rs
ludo-c: Indeed, `PathBuf` is better, I changed it, thank you !

If you are ok with the last commit I will squash it into the "Fix process iterator with a custom root" one.

ludo-c: I squashed it !

CI is broken because of the version in Cargo.toml, I don't think I can change that.

tatref: There are multiple issues in CI atm, but not related to your changes

- rust 1.48 has an issue with the manifest format
- stable and nightly have an issue with sys::kernel::random::tests::test_poolsize. It seems related to a recent change in the kernel, see https://unix.stackexchange.com/questions/704737/kernel-5-10-119-caused-the-values-of-proc-sys-kernel-random-entropy-avail-and-p

However, I don't know if the change of failure to anyhow should be handled in this PR (I'm not the repo owner, though)

ludo-c: As "failure" is used only in test and is deprecated, I thought it worst nothing to add it here.

If it is a problem I will of course remove it from the PR. The main issue is the real important thing to fix.

Thank you for your review !

tatref:
I submitted a PR to fix CI https://github.com/eminence/procfs/pull/205

There's still some issue with a dependency, but it should be fixed soon

src/process/mod.rs
```diff
@@ -1538,8 +1542,9 @@ impl std::iter::Iterator for ProcessesIter {
             match self.inner.next() {
                 Some(Ok(entry)) => {
                     if let Ok(pid) = i32::from_str(&entry.file_name().to_string_lossy()) {
-                        if let Ok(proc) = Process::new(pid) {
-                            break Some(Ok(proc));
+                        match Process::new_with_root(self.root.join(pid.to_string())) {
+                            Ok(proc) => break Some(Ok(proc)),
+                            Err(e) => break Some(Err(e)),
```
eminence, line 1547: The documented behavior of the `all_processes` function is that `Process` objects that fail to be constructed won't be returned in the iterator.  Your code here would modify that, and I don't think I want to do that right now.

If you have a use-case where you want to know about these failures, please let me know.

src/sys/kernel/random.rs
```diff
@@ -73,7 +73,7 @@ mod tests {
         // The kernel support section in the root lib.rs file says that we only aim to support >= 2.6 kernels,
         // so only test that case
         let poolsize = poolsize().unwrap();
-        assert!(poolsize == 4096)
+        assert_eq!(poolsize, 4096)
```
eminence, line 76: This will conflict with #205.  That's OK, just FYI

src/lib.rs
```diff
@@ -1180,6 +1180,20 @@ mod tests {
         println!("{:?}", load);
     }
 
+    #[test]
+    fn test_loadavg_from_reader() -> ProcResult<()> {
+        let file_content = String::from(r#"2.63 1.00 1.42 3/4280 2496732"#);
+        let load_average = LoadAverage::from_reader(file_content.as_bytes())?;
```
eminence, line 1186:
```suggestion
        let load_average = LoadAverage::from_reader("2.63 1.00 1.42 3/4280 2496732".as_bytes())?;
```

eminence (changes requested):
Thanks for this fix @ludo-c !    The bug fix looks good to me in general, but I did leave two comments on things that need changing before merging.

Cargo.toml
```diff
@@ -30,7 +30,7 @@ serde = { version = "1.0", features = ["derive"], optional = true }
 [dev-dependencies]
 criterion = "0.3"
 procinfo = "0.4.2"
-failure = "0.1"
+anyhow = "1.0"
```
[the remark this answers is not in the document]
ludo-c: I did this because `failure` has been marked as deprecated : https://github.com/rust-lang-deprecated/failure

I remove it from this PR.

src/sys/kernel/random.rs
ludo-c: I remove it too.

eminence:
(Also, please rebase onto the last master branch)

src/process/mod.rs
ludo-c: It was useful only to find why I didn't have all the processes I expected (in my case `/proc` has less process than `/proc-full`)
With this fix it is not needed anymore, I remove it.

eminence (approved):

eminence:
Thanks again for the fix!

ludo-c:
Hi @eminence .
As there have been some fixes, would it be possible to bump a 0.15 version please ?

Thank you !

eminence:
Sure, we can get a new release out in a few days

PR #209, opened 2022-10-10
Fix memfd fd target

trtt:
Memfd case is unreachable since the whole match block is for `!s.starts_with('/')`.


eminence:
Nice find, thank you!

PR #210, opened 2022-10-14
Disabling default features on the `chrono` crate

Will-Low:
# Situation
The `chrono` crate depends on an outdated version of the `time` crate, which has a known vulnerability. `chrono` apparently does not use the vulnerable element, but it has breaking backwards compatibility issues and will not be removed until they upgrade the major version. There is a workaround suggested in https://github.com/chronotope/chrono/issues/602#issuecomment-1225331424. I've compiled it myself and confirm the vulnerable crate no longer is a dependency in `Cargo.lock`.

# Target
Remove the vulnerability to prevent vulnerability-tool noise.

# Proposal
In accordance with the workaround, disable the default features of `chrono` and instead replace it with the "clock" feature, which does not compile the vulnerable crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ serde1 = ["serde"]
 rustix = { version = "0.35.6", features = ["fs", "process", "param", "thread"] }
 bitflags = "1.2"
 lazy_static = "1.0.2"
-chrono = {version = "0.4.20", optional = true }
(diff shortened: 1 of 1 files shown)


eminence:
Hi, sorry for the delay in sending a reply.  Thanks for this PR.   My understanding is that this change is fully semver compatible, since it's only removing features that `procfs` never used in the first place.  Is that right?

eminence:
This change still looks good to me, so I'm going ahead an merging this.

Issue #126, opened 2021-05-05
Allow reading from specific pid's /net/dev


zz85:
Seems like it should be an easy change to extend reading from /net/dev to specific processes

https://github.com/eminence/procfs/blob/d8f16ddc97413ab2fb38d23149be188ad0b750a7/src/net.rs#L619

eg.

`/proc/self/net/dev`

`/proc/<pid>/net/dev`

eminence:
Yup, agreed.  Note that `/proc/net` is normally just a symlink to `/proc/self/net`.

I'm curious: what's your use-case to read the `/proc/net/dev` of a specific process?

zz85:
I was planning to monitor bandwidth (bytes send/recv) usage for processes, and was thinking that /proc/net would be a place i could implemented it with...

PR #211, opened 2022-11-14
read dev and tcp files with specific pid

shippomx:
I have to write some kata-containers metrics interface, which will use the specific /proc/{pid}/net/tcp file. And some one has mentioned this problem one yeas ago. [Allow reading from specific pid's /net/dev #126](https://github.com/eminence/procfs/issues/126)

### DIFF
--- a/src/net.rs
+++ b/src/net.rs
@@ -348,6 +348,20 @@ pub fn tcp6() -> ProcResult<Vec<TcpNetEntry>> {
     read_tcp_table(BufReader::new(file))
 }
 
+/// Reads the specific tcp socket table
+pub fn tcp_with_pid(pid: i32) -> ProcResult<Vec<TcpNetEntry>> {
+    let file = FileWrapper::open(format!("/proc/{}/net/tcp", pid))?;
+
+    read_tcp_table(BufReader::new(file))
+}
+
+/// Reads the specific tcp6 socket table
+pub fn tcp6_with_pid(pid: i32) -> ProcResult<Vec<TcpNetEntry>> {
+    let file = FileWrapper::open(format!("/proc/{}/net/tcp6", pid))?;
+
+    read_tcp_table(BufReader::new(file))
+}
+
 /// Reads the udp socket table
 pub fn udp() -> ProcResult<Vec<UdpNetEntry>> {
     let file = FileWrapper::open("/proc/net/udp")?;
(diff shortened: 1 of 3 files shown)


src/net.rs
```diff
@@ -59,6 +59,7 @@ use byteorder::{ByteOrder, NativeEndian, NetworkEndian};
 use std::io::{BufRead, BufReader, Read};
 use std::net::{Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6};
 use std::{path::PathBuf, str::FromStr};
+use libc::pid_t;
```
eminence, line 62: I would to not add the `libc` dep, just for the `pid_t` type.  Can we just use an `i32` (which is what we also use in the `Process` module) ?

src/net.rs
```diff
@@ -641,6 +661,17 @@ pub fn dev_status() -> ProcResult<HashMap<String, DeviceStatus>> {
     Ok(map)
 }
 
+/// Returns basic network device statistics for all interfaces
+///
+/// This data is from the `/proc/{pid}/net/dev` file.
+///
+/// For an example, see the [interface_stats.rs](https://github.com/eminence/procfs/tree/master/examples)
+/// example in the source repo.
```
eminence, line 669: This comment is not accurate, since the interface_stats example doesn't use `dev_status_with_pid`

src/net.rs
```diff
@@ -348,6 +349,20 @@ pub fn tcp6() -> ProcResult<Vec<TcpNetEntry>> {
     read_tcp_table(BufReader::new(file))
 }
 
+/// Reads the specific tcp socket table
+pub fn tcp_with_pid(pid: pid_t) -> ProcResult<Vec<TcpNetEntry>> {
+    let file = FileWrapper::open(format!("/proc/{}/net/tcp", pid))?;
+
+    read_tcp_table(BufReader::new(file))
+}
```
eminence, line 357: How is this different than `Process::tcp()` ?

eminence:
Hi @shippomx, please let me know what you think about the comments I left.

src/net.rs
shippomx: Sorry for late, I think you are right. There is no need to import a new package for just using the pid_t type. I will renew this pull request later.

PR #213, opened 2022-11-23
Add smaps_rollup to support.md.

TaborKelly:


TaborKelly:
Thanks for the great project! I might work on code to support `smaps_rollup` soon.

PR #214, opened 2022-11-23
Implement smaps_rollup

TaborKelly:
smaps_rollup is a lot faster than adding up all the smaps but is only available in newer kernels.


eminence:
LGTM, thanks!

TaborKelly:
@eminence, thanks again for the quick help to get this merged in. FYI, we used it to write and then open source this simple but very helpful tool: https://gitlab.com/racepointenergy/tools/rps

PR #215, opened 2022-11-28
Update to rustix 0.36.

sunfishcode:
The main change here is the change to io-lifetimes 1.0, which uses the `OwnedFd` etc. from std, on Rust versions where it's available.

This version also drops the `FromFd` trait, because std uses `From<OwnedFd>` instead.


eminence:
Thanks!

PR #217, opened 2022-11-30
pagemap / PFN example

tatref:
This example print physical memory location for each virtual memory mapping.

This requires root to work properly.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -142,3 +142,13 @@ Lists all processes as a tree. Sub-processes will be hierarchically ordered bene
 ...
 ```
 
+## pfn.rs
+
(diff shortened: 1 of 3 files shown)


examples/README.md
```diff
@@ -142,3 +142,13 @@ Lists all processes as a tree. Sub-processes will be hierarchically ordered bene
 ...
 ```
 
+## pfn.rs
+
+List memory mapping, and physical address for each virtual address. Must be run as root, see [pagemap.txt](https://github.com/torvalds/linux/blob/v4.9/Documentation/vm/pagemap.txt)
```
eminence, line 147:
```suggestion
List memory mapping, and physical address for each virtual address. Must be run as root, see [pagemap.txt](https://www.kernel.org/doc/Documentation/vm/pagemap.txt)
```

eminence (commented):
LGTM, with just one minor URL adjustment.

PR #218, opened 2022-11-30
read dev and tcp files with specific pid #

shippomx:
I have to write some kata-containers metrics interface, which will use the specific /proc/{pid}/net/tcp file. And some one has mentioned this problem one yeas ago. https://github.com/eminence/procfs/issues/126.

for some mistake, i had add `libc` package to reference `pid_t` type, but this pr i removed it, and test this function in `examples/interface_stats.rs`

### DIFF
--- a/src/net.rs
+++ b/src/net.rs
@@ -348,6 +348,20 @@ pub fn tcp6() -> ProcResult<Vec<TcpNetEntry>> {
     read_tcp_table(BufReader::new(file))
 }
 
+/// Reads the specific tcp socket table
+pub fn tcp_with_pid(pid: i32) -> ProcResult<Vec<TcpNetEntry>> {
+    let file = FileWrapper::open(format!("/proc/{}/net/tcp", pid))?;
+
+    read_tcp_table(BufReader::new(file))
+}
+
+/// Reads the specific tcp6 socket table
+pub fn tcp6_with_pid(pid: i32) -> ProcResult<Vec<TcpNetEntry>> {
+    let file = FileWrapper::open(format!("/proc/{}/net/tcp6", pid))?;
+
+    read_tcp_table(BufReader::new(file))
+}
+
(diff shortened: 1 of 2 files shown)


src/net.rs
```diff
@@ -348,6 +348,20 @@ pub fn tcp6() -> ProcResult<Vec<TcpNetEntry>> {
     read_tcp_table(BufReader::new(file))
 }
 
+/// Reads the specific tcp socket table
+pub fn tcp_with_pid(pid: i32) -> ProcResult<Vec<TcpNetEntry>> {
+    let file = FileWrapper::open(format!("/proc/{}/net/tcp", pid))?;
+
+    read_tcp_table(BufReader::new(file))
+}
```
eminence, line 356: I'm still not sure how this new function is different from [Process::tcp()](https://docs.rs/procfs/0.14.1/procfs/process/struct.Process.html#method.tcp)

src/net.rs
```diff
@@ -641,6 +660,17 @@ pub fn dev_status() -> ProcResult<HashMap<String, DeviceStatus>> {
...
+/// Returns basic network device statistics for all interfaces
+///
+/// This data is from the `/proc/{pid}/net/dev` file.
+///
+/// For an example, see the [interface_stats.rs](https://github.com/eminence/procfs/tree/master/examples)
+/// example in the source repo.
+pub fn dev_status_with_pid(pid: i32) -> ProcResult<HashMap<String, DeviceStatus>> {
+    let file = FileWrapper::open(format!("/proc/{}/net/dev", pid))?;
+    read_dev_status(file)
+}
```
eminence, line 672: This function should be part of the `Process` struct

eminence:
hi @shippomx -- I just merged #226, which supersedes your pull request, so I'm going to close this.  Thank you though for your PR.

PR #220, opened 2022-12-07
fix chrono::Local::timestamp deprecated in chrono 0.4.23

tatref:
I'm not 100% sure about the `expect!` macro.

Tell me if it's good


eminence:
This looks right to me.  Our timestamp should never be invalid, and according to the `timestamp_opt` docs, it will never return an `Ambiguous` value

PR #231, opened 2023-01-08
parse uid for /proc/net/{tcp,udp}

trinity-1686a:
Hi,

I was evaluating this crate for [my use case](https://gitlab.torproject.org/tpo/core/onionmasq/-/issues/16), and it happens to skip over the exact bit of data I'm interested in.

I also added some `#[non_exhaustive]` on structures where some columns are skipped over, so that parsing these in the future isn't a breaking change semver-wise

### DIFF
--- a/src/net.rs
+++ b/src/net.rs
@@ -173,6 +173,7 @@ impl UnixState {
 
 /// An entry in the TCP socket table
 #[derive(Debug, Clone)]
+#[non_exhaustive]
 #[cfg_attr(feature = "serde1", derive(Serialize, Deserialize))]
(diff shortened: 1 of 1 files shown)


trinity-1686a:
hum, CI is unhappy but it looks like it's not related to my changes. Sleeping 30ms might not always be enough in `test_proc_alive`

eminence:
Yeah, I think that 30ms sleep needs to be bumped up.  The test passed on re-run.  Thanks for the PR, I'm not sure why I ignored UID in the first place.  This LGTM

PR #232, opened 2023-01-11
add /proc/kpagecount

tatref:


src/kpagecount.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /// Get a parser from custom `/proc`
+    ///
+    /// Return `Err` if process is not running as root or don't have CAP_SYS_ADMIN
+    pub fn from_custom_root<P: AsRef<Path>>(root: P) -> ProcResult<Self> {
+        let mut path = root.as_ref().to_path_buf();
+        path.push("kpagecount");
+
+        let reader = File::open(&path)?;
```
eminence, line 29: Can this use `FileWrapper::open(..)` instead?

tatref: I don't know why I did it differently than kpageflags,

I refactored so the code for kpageflags and kpagecount is now similar.

tatref:
At the moment, `get_count_in_range` is performing a lot of reads. I'll benchmark and try to perform a single read for the whole Vec if it's worth it.

Also, I copy pasted some code from process::pagemap, but the unbounded ranges don't make any sense (same for page map). We should raise an error if this is the case.

eminence:
> I'll benchmark and try to perform a single read for the whole Vec if it's worth it.

Do you want to hold off on the merge until this is done?  Or should we merge this now, and iterate if necessary in future pull requests?

tatref:
Did some testing replacing the multiple reads in the loop by a single read, but program now runs slower... I guess this is not the main bottleneck.

I think this can be merged.

One future enhancement would be to replace u64 by a new type like Pfn(u64)

PR #234, opened 2023-01-14
Enable oppportunistic fd counting fast path

bobrik:
Existing slow path takes ~52000us of CPU time on my laptop to count 100k open files. Compare this to just 13us for the baseline, when no extra files are open by a program:

```
fd count = 7, elapsed = 13 us
fd count = 100007, elapsed = 51863 us
```

Adding fastpath from Linux v6.2 makes it fast:

```
fd count = 4, elapsed = 2 us
fd count = 100004, elapsed = 8 us
```

This is before taking in account any lock contention effects in the kernel if you try to count files from multiple threads concurrently, which makes the slow path even slower, burning a lot more CPU in the process. See:

* https://github.com/torvalds/linux/commit/f1f1f2569901

I used the following code to benchmark:

```rust
use procfs::process::Process;
use std::{fs::File, time::Instant};

fn main() {
    let myself = Process::myself().unwrap();

    let started = Instant::now();
    let count = myself.fd_count().unwrap();
    let elapsed_us = started.elapsed().as_micros();

    eprintln!("fd count = {}, elapsed = {} us", count, elapsed_us);

    let files = (0..100_000)
        .map(|_| File::open("/etc/hosts").unwrap())
        .collect::<Vec<_>>();

    let started = Instant::now();
    let count = myself.fd_count().unwrap();
    let elapsed_us = started.elapsed().as_micros();

    eprintln!("fd count = {}, elapsed = {} us", count, elapsed_us);

    drop(files);
}
```

See also: https://github.com/prometheus/procfs/pull/486

### DIFF
--- a/src/process/mod.rs
+++ b/src/process/mod.rs
@@ -1084,6 +1084,15 @@ impl Process {
     ///
     /// Calling this function is more efficient than calling `fd().unwrap().count()`
     pub fn fd_count(&self) -> ProcResult<usize> {
+        // Use fast path if available (Linux v6.2): https://github.com/torvalds/linux/commit/f1f1f2569901
+        let stat = wrap_io_error!(
+            self.root.join("fd"),
+            rustix::fs::statat(&self.fd, "fd", AtFlags::SYMLINK_NOFOLLOW)
+        )?;
+        if stat.st_size > 0 {
+            return Ok(stat.st_size as usize);
+        }
+
         let fds = wrap_io_error!(
             self.root.join("fd"),
             rustix::fs::openat(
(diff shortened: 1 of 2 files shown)


eminence:
Looks nice, thanks.  I just realized that we don't have any tests for `fd_count()`... would you mind adding a small commit that calls `fd_count` -- probably best in `test_all()` or `test_proc_fds()`

bobrik:
Sure, I added a basic test to make sure that the count increments and decrements as expected. It needs to run single threaded to avoid picking up fds from other threads. Let me know if you had something else in mind.

eminence:
That's perfect, thank you!

PR #237, opened 2023-01-23
Refactor and expose memory map processing.

### LINKED ISSUE #236: Expose `/proc/<pid>/maps` parsing (and others)
We'd like to expose the parsing of `/proc/<pid>/maps` in the external interface to be able to use it on stored memory map information in [rust-minidump](https://github.com/rust-minidump/rust-minidump). We currently parse the maps ourselves, however we are interested in using a separate crate so that we may re-use the parsing amongst multiple crates that we have. Since `procfs` basically already has this and much more, we figured it'd be great if we could use it instead. We also have other needs for parsing various `procfs` information so we would likely use it in the future for those purposes, too.

We're content with contributing this API exposition, however we wanted to check with the primary maintainer(s) that this is a desirable API to be added to the crate. Specifically, the `procfs` crate currently describes itself as "... an interface to the proc pseudo-filesystem on linux", and if we were to add functions which can operate on raw/stored `procfs` information, they would not _technically_ be interfacing with the proc pseudo-filesystem at all (e.g. parsing the stored content of the `maps` file doesn't involve interacting directly with any mounted `procfs`). We don't expect this would be a problem, but it's worth noting the distinction.

Thanks!

afranchuk:
A new MemoryMaps struct is added, which parses all of `maps`, `smaps`, and `smaps_rollup`.

Closes #236.

### DIFF
--- a/src/process/mod.rs
+++ b/src/process/mod.rs
@@ -566,15 +566,23 @@ pub struct MemoryMaps {
 }
 
 impl MemoryMaps {
+    /// Read a [MemoryMaps] from the given byte source.
+    ///
+    /// The data should be formatted according to procfs /proc/pid/{maps,smaps,smaps_rollup}.
     pub fn from_reader<R: io::Read>(r: R) -> ProcResult<Self> {
         Self::read(r, None)
     }
 
+    /// Read a [MemoryMaps] from the given path.
+    ///
+    /// The file data should be formatted according to procfs
+    /// /proc/pid/{maps,smaps,smaps_rollup}.
     pub fn from_path<P: AsRef<Path>>(path: P) -> ProcResult<Self> {
         let file = FileWrapper::open(path.as_ref())?;
         Self::read(file, Some(path.as_ref()))
     }
 
+    /// Return an iterator over [MemoryMap].
     pub fn iter(&self) -> std::slice::Iter<MemoryMap> {
         self.memory_maps.iter()
     }
@@ -1143,7 +1151,7 @@ impl Process {
     /// Return a list of the currently mapped memory regions and their access permissions, based on
     /// the `/proc/pid/maps` file.
     pub fn maps(&self) -> ProcResult<MemoryMaps> {
-        MemoryMaps::from_path(self.root.join("maps"))
+        MemoryMaps::from_reader(FileWrapper::open_at(&self.root, &self.fd, "maps")?)
     }
 
     /// Returns a list of currently mapped memory regions and verbose information about them,
@@ -1151,16 +1159,14 @@ impl Process {
     ///
     /// (since Linux 2.6.14 and requires CONFIG_PROG_PAGE_MONITOR)
     pub fn smaps(&self) -> ProcResult<MemoryMaps> {
-        MemoryMaps::from_path(self.root.join("smaps"))
+        MemoryMaps::from_reader(FileWrapper::open_at(&self.root, &self.fd, "smaps")?)
     }
 
     /// This is the sum of all the smaps data but it is much more performant to get it this way.
     ///
     /// Since 4.14 and requires CONFIG_PROC_PAGE_MONITOR.
     pub fn smaps_rollup(&self) -> ProcResult<SmapsRollup> {
-        let file = FileWrapper::open_at(&self.root, &self.fd, "smaps_rollup")?;
-
-        SmapsRollup::from_reader(file)
+        SmapsRollup::from_reader(FileWrapper::open_at(&self.root, &self.fd, "smaps_rollup")?)
     }
 
     /// Returns a struct that can be used to access information in the `/proc/pid/pagemap` file.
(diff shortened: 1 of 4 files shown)


eminence:
If you rebase onto the latest master branch, the CI issues will clear up (see #239)

afranchuk:
Looks good now!

afranchuk:
Hold up, forgot to add the permissions parsing (it _could_ be a separate PR but there's probably no need).

afranchuk:
That should do it :)

afranchuk:
@eminence does this look good now?

eminence:
Sorry about the delay on my side.  I'm going to take a look Very Soon Now™

src/process/smaps_rollup.rs
```diff
@@ -1,58 +1,14 @@
...
-                let k = k.trim_end_matches(':');
-
-                memory_map_data.map.insert(k.into(), v * size_multiplier);
-            }
-        }
-        Ok(SmapsRollup {
-            memory_map,
-            memory_map_data,
-        })
+        MemoryMaps::from_reader(r).map(|m| SmapsRollup { memory_map_rollup: m })
```
eminence, line 12: It's always nice to see this type of cleanup 👍

src/process/mod.rs
```diff
@@ -983,85 +1134,16 @@ impl Process {
...
-        let mut vec = Vec::new();
-
-        for line in reader.lines() {
-            let line = line.map_err(|_| ProcError::Incomplete(Some(self.root.join("maps"))))?;
-            vec.push(MemoryMap::from_line(&line)?);
-        }
-
-        Ok(vec)
+    pub fn maps(&self) -> ProcResult<MemoryMaps> {
+        MemoryMaps::from_path(self.root.join("maps"))
```
eminence, line 1138: Instead of loading by path, can we use the same `open_at` technique that the old version has?  Something like:

```rust
MemoryMaps::from_reader(FileWrapper::open_at(&self.root, &self.fd, "maps")?)
```

src/process/mod.rs
```diff
@@ -510,15 +557,125 @@ impl MMapPath {
     }
 }
 
-/// Represents an entry in a `/proc/<pid>/maps` file.
+/// Represents all entries in a `/proc/<pid>/maps` or `/proc/<pid>/smaps` file.
+#[derive(Debug, PartialEq, Eq, Clone)]
+#[cfg_attr(feature = "serde1", derive(Serialize, Deserialize))]
+pub struct MemoryMaps {
```
eminence, line 563: It might make sense to declare this struct `#[non_exhaustive]` so we can add new fields in the future in a backwards-compatible way

src/process/mod.rs
```diff
@@ -529,20 +686,13 @@ pub struct MemoryMap {
     /// BSS (uninitialized data).
     pub inode: u64,
     pub pathname: MMapPath,
+    /// Memory mapping extension information, populated when parsing `/proc/<pid>/smaps`.
+    ///
+    /// The members will be `Default::default()` (empty/none) when the information isn't available.
+    pub extension: MMapExtension,
```
eminence, line 692: I'm wondering if anyone will have a usecase where they want to quickly know if the `extension` data is empty/none.  Would an `Option<MMapExtension>` type be better?

No need to change anything now, just starting the brainstorming process.

eminence (commented):
LGTM, just a few comments.  Thanks for working on this

src/process/mod.rs
afranchuk: I couldn't really see any advantage to using `open_at`, but if you prefer that I can change it.

src/process/mod.rs
afranchuk: I considered that, and intended to add an `is_empty` method but forgot. Though I can't actually think of a convincing use case, only because the typical "eliding unnecessary code" use case is pretty easy to check by accessing the inner members (e.g. `map`). I chose not to add `Option` because I thought it would be more ergonomic for more general code to simply query the inner members rather than having to match and remove the `Option` layer in some way.

eminence: I'm liking the `is_empty` solution 👍

src/process/mod.rs
eminence: `open_at` helps in case there's pid reuse.  The scenario looks like:

You create a `Process` object for pid `123`.  Internally, we open the `/proc/123` directory.  Some time later, this process exists, but our `Process` object still exists.    If another process is later started that also has pid `123`, then opening `/proc/123/maps` by path will be successful.  But this is problematic because we've now reading maps info for a totally different process.

With `open_at` on the previously opened `/proc/123` directory, trying to open the maps file will result in an error indicating that the process is no longer running (which is what we want)

afranchuk: Understood! Made the change.

PR #238, opened 2023-01-23
Add new `/proc/meminfo` fields.

afranchuk:
`Zswap`, `Zswapped`, and `SecPageTables` are added.

### DIFF
--- a/src/meminfo.rs
+++ b/src/meminfo.rs
@@ -145,6 +145,11 @@ pub struct Meminfo {
     ///
     /// (since Linux 2.6.18)
     pub page_tables: Option<u64>,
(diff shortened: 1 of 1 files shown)


src/meminfo.rs
```diff
@@ -145,6 +145,11 @@ pub struct Meminfo {
     ///
     /// (since Linux 2.6.18)
     pub page_tables: Option<u64>,
+    /// Amount of memory allocated for seconary page tables. This currently includes KVM mmu
+    /// allocaitons on x86 and arm64.
```
eminence, line 149:
```suggestion
    /// allocations on x86 and arm64.
```

PR #241, opened 2023-02-02
fix from_process_at return error in old kernel

qq906907952:
In the function "FDInfo::from_process_at" it use  "rustix::fs::openat" to open file like "/proc/{pid}/fd/x"
which will encounter the 'Os { code: 40, kind: FilesystemLoop, message: "Symbolic link loop" }' error in old version kernel 
(test in  2.6.32-696.16.1.el6.x86_64, i guess the some flag not handle correct or not support in old kernel version, espcially SYMLINK_NOFOLLOW)

when i remove  SYMLINK_NOFOLLOW flag in openat ,  "rustix::fs::openat" raise 'Os { code: 6, kind: Uncategorized, message: "No such device or address" }' error

it also encounter 'Os { code: 22, kind: InvalidInput, message: "Invalid argument" }'  when get metadata use `rustix::fs::statat` with flags.

so i remove the "rustix::fs::openat" and remove all  AtFlags  in "rustix::fs::statat". only use "rustix::fs::readlinkat" and "rustix::fs::statat"  to get the fd type and metadata

also remove the unused field `root` in struct FDsIter

it work in old kernel (i only test in  2.6.32-696.16.1.el6.x86_64) and new kernel

### DIFF
--- a/src/process/mod.rs
+++ b/src/process/mod.rs
@@ -890,26 +890,10 @@ impl FDInfo {
     /// Gets a file descriptor from a directory fd and a path relative to it.
     ///
     /// `base` is the path to the directory fd, and is used for error messages.
-    fn from_process_at<P: AsRef<Path>, Q: AsRef<Path>>(
-        base: P,
-        dirfd: BorrowedFd,
-        path: Q,
-        fd: i32,
-    ) -> ProcResult<Self> {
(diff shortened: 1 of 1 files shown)


eminence:
Hi, thanks for the bug report, and for the potential fix.  I would like to be able to support 2.6 kernels, but getting ahold of one for testing is a bit challenging.  I need to do some more research to understand the impact of removing `SYMLINK_NOFOLLOW` and `EMPTY_PATH` from the calls to `statat`

PR #247, opened 2023-02-14
Don't error if the oppportunistic statat fails in fd_count

eminence:
It was reported (#241) that old kernels (2.6) will return a failure if SYMLINK_NOFOLLOW is used, so lets check for errors here.


eminence:
@bobrik  does this look OK to you?

bobrik:
I think we can also just not pass any flags (pass `AtFlags::empty()`) and that would work too and be easier to follow.

eminence:
Ah, good point.  I somehow had it in my head that the `SYMLINK_NOFOLLOW` part was relevant here, but that's not true.  So `AtFlags::empty()` sounds like the way to go here.  Thanks

bobrik (approved):

Issue #245, opened 2023-02-13
PID should be u32


tatref:
Hi,

PID is currently i32:

https://github.com/eminence/procfs/blob/3d5c9fa43b4e9f4401de4de71328401efe9ecd1a/src/process/mod.rs#L957

However, in stdlib, PID is u32: https://doc.rust-lang.org/std/process/fn.id.html

Is there any reason for this?

eminence:
I think it's because there are a few procfs fields that can return `-1` (if the pid isn't available or something).   I can't name these fields off the top of my head, but it would be good to find them and document them.

eminence:
Based on the notes in #250, we're keeping pid as `i32`

PR #250, opened 2023-02-14
Convert PID and TID to u32

tatref:
See issue #245

I tried to convert all the i32 fields to u32 for Process and Task.
I only had to mess a bit with the display code for the example lslocks.

Cargo check and test are OK, and I tested all the examples, everything works as expected

Feel free to discard this PR if you prefer to keep the current i32.

### DIFF
--- a/examples/dump.rs
+++ b/examples/dump.rs
@@ -1,7 +1,7 @@
 extern crate procfs;
 
 fn main() {
-    let pid = std::env::args().nth(1).and_then(|s| s.parse::<i32>().ok());
+    let pid = std::env::args().nth(1).and_then(|s| s.parse::<u32>().ok());
 
     let prc = if let Some(pid) = pid {
         println!("Info for pid={}", pid);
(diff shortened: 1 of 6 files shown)


eminence:
I have no objection to this change, but I want to do some digging to quell the niggling feeling in the back of my mind that I had (once upon a time) a good reason to pick `i32`

rust1248:
Maybe because libc `pid_t` is i32?

eminence:
Also maybe because the [`proc.5`](https://man7.org/linux/man-pages/man5/proc.5.html) page says that when parsing `/proc/pid/stat`, the PID field should be parsed with `%d` (which is a signed int).

tatref:
You are correct!
Seems like this is already known: https://internals.rust-lang.org/t/std-child-pid-type-compatibility/10473/1

I guess this is not worth pursuing

eminence:
Ok, let's keep the current `i32` for now.   Thanks for testing this, though.

PR #251, opened 2023-02-15
Change Stat::comm documentation

### LINKED ISSUE #249: Confusing documentation for Stat::comm
The documentation [says](https://docs.rs/procfs/latest/procfs/process/struct.Stat.html#structfield.comm):

```text
The filename of the executable, in parentheses.

This is visible whether or not the executable is swapped out.

Note that if the actual comm field contains invalid UTF-8 characters, they will be replaced here by the U+FFFD replacement character.
```

What I understood was that the parentheses were included in the field. I was surprised to see it without.

rust1248:
Close #249


eminence:
Thank you!

PR #254, opened 2023-02-19
Implement some standard traits

tatref:
I implemented a few standard traits, but there's probably others.

For `Shm`, some field like `rss` or `atime` are not really part of the indentity of `Shm`, so I left them out

### DIFF
--- a/procfs-core/src/process/pagemap.rs
+++ b/procfs-core/src/process/pagemap.rs
@@ -98,7 +98,7 @@ impl fmt::LowerHex for Pfn {
 }
 
 /// Represents a page table entry in `/proc/<pid>/pagemap`.
-#[derive(Debug)]
+#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
 pub enum PageInfo {
     /// Entry referring to a memory page
     MemoryPage(MemoryPageFlags),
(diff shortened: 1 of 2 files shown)


tatref:
Converted to draft, as the `Hash` for `Shm` is probably plain wrong if some fields are not part of the hash

eminence:
The derived traits seem OK to me, but I agree that the custom impl for `Shm` seems a little suspect.  For me, the primary purpose to have a type impl `Hash` is to be able to use it has a key in a `HashMap`.  I guess it's not inconceivable that you'd want to do this.  Did you have a specific usecase for implementing `Hash` on `Shm` ?

tatref:
I'm working on a tool for memory diagnostics.
I want to store a list of physical pages used by each shared memory segment, so `HashMap<Shm, Vec<Pfn>>` makes a lot of sense.

I'll rework the PR to remove the custom impl stuff.

eminence:
I'm not saying that `procfs` can't implement these, but could you put `Shm` in your own local wrapper type, and impl `Hash` and `PartialEq` according to the rules that make sense for your tool?

tatref:
After some thinking, I believe that the custom `Hash` and `PartialEq` that I initially implemented is plain wrong.
I switched to using derive

Tell my if it's good for you

tatref:
I rebased from master

Issue #255, opened 2023-02-23
Support (limited) use on other platforms


afranchuk:
`procfs` only compiles on linux and android. For our use case, we only use `procfs` for parsing files which were previously stored; there's no need to limit platforms in this case. We'd like to be able to use this subset of cross-platform-capable features, but currently cannot.

I began some changes to allow support on other platforms, however there are some design decisions for which I wanted feedback. FWIW, there are only a handful of things that aren't macos-compatible currently. There are more things that aren't windows-compatible, though mostly it's anything using `rustix::fs`, otherwise most of the incompatibilities have known fixes (e.g. `OsStr` conversions). It might actually be feasible to only exclude these (rather than wholesale excluding most stuff).

There are a few ways to achieve this. Below I enumerate them, where the term "local-procfs" is used to designate systems which (should) have a `procfs` mounted. It could just as easily be "native-procfs" or something else.

#### Options
1. Change `build.rs` to provide a `cfg(local-procfs)` or an equivalent feature flag rather than failing the build.
2. Add a feature flag for local-procfs, remove `build.rs`, and enforce with a `compile_error!` that it isn't used on unsupported platforms. TBH the crate failing to compile otherwise would basically be error enough, though having a better error message is nice.
3. Separate the `procfs` type definitions and basic parsers to another crate.

#### Notes
* (1) is a bit inflexible only because some people might want to use the crate only for cross-platform purposes, in which case there's no reason to compile all the other stuff when targeting linux/android.
* (1) and (2) are both a bit cumbersome to implement, because unfortunately there's no way to tell rustc to compile _only_ stuff with a particular `cfg` designation, so you have to go through _all_ existing code and tag it (and remember to do so for future code, too). This could be eased significantly by separating the code into distinct modules (so that you can simply include/exclude whole sets of implementations), but that separation would come at the cost of splitting many existing files in two (either in the filesystem or within the files with `mod`s).
* (3) is essentially similar to the end of the last note (making distinct modules for the support), though it's a bit more formal in separation.

If this isn't something that's desirable for the `procfs` crate, we may have to use something else (though perhaps we can work around it by changing how some other tools store data). However we do have a decent amount of things to read from `procfs` files, so it'd be a shame to not be able to utilize the code you've created.

eminence:
[...]

PR #257, opened 2023-03-03
Implementation of a split crate scheme.

afranchuk:
This splits the crates as a demonstration of what that might look like.

cc #255


eminence:
Thanks for starting work on this!

For stuff like `fn split_into_num()` and `trait FromStrRadix` (which is used in both `procfs-core` and `procfs`), what did you mean by moving them into a module?   I think I'm OK making these previously private stuff be moved into `procfs-core` and made public.

Github helpfully gives me access to your `split-crates` branch, so we can collaborate there, if you want; just let me know what you're working on, so I won't step on your toes.

eminence:
Ah, I see.  For now, let's just make them public in `procfs-core`.  Later we can consider applying `#[doc(hidden)]` and/or adding some comment like "this function isn't actually part of the public API".  Putting them in an `internal` module would also help communicate that these are internal APIs, but this PR is going to be pretty big anyway, so let's iterate on this topic later

afranchuk:
@eminence Take a look at what I've got so far. More stuff could be moved to `procfs-core`, but the current state demonstrates how the code is split (for instance see how `process` works now, that is in a "finished" state). In particular, see the traits that were added to `procfs-core/src/lib.rs` and `procfs/src/lib.rs`, which make a consistent interface for things.

This includes allowing the (new) `SystemInfo` to be provided explicitly to `procfs-core`, which `procfs` will default to the local system. That was originally a struct (the same as `ExplicitSystemInfo`), but I made it a trait object so that the `ProcResult` in functions were one-to-one as before (otherwise, e.g., you wouldn't be able to get `ticks_per_second` if `boot_time_secs` was unavailable for whatever reason, since an error would be returned.

_Types_ which need `SystemInfo` to be created get it through a set of parallel traits.

_Methods_ which need `SystemInfo` get it by returning an `impl WithSystemInfo`. I chose this approach (rather than creating extension traits for the types which need an alternate method in `procfs`) because it was more uniform and allowed the original method name to remain the same (otherwise for user convenience you'd want the method names to be different to avoid requiring the user to disambiguate when calling them).

eminence:
(Just dropping a quick note to say that I haven't forgotten about this PR)

afranchuk:
No problem, I know it's a lot to take in (and still isn't complete in moving stuff to `procfs-core`, though it's not far)!

afranchuk:
Also for such a large changeset I recommend simply looking at the files. The diff view is way more distracting than helpful (I wish they at least supported side-by-side diffs).

eminence:
I just pushed 2 new commits to your branch: one updates github CI, and the other removes some problematic doc comments (which were trying to link to `procfs` stuff from `procfs_core`).  Let's see if this is enough to turn CI green

sunfishcode:
Yes, it looks like the nightly failure is due to https://github.com/rust-lang/rust/issues/109424, which happens to hit rustix.

afranchuk:
Aside from CI blinking lights, how does the crate organization look? Is this a direction you want to take the project?

afranchuk:
Yeah, the `procfs-core` crate meets my needs. Using a semver tool is a good idea to sanity check the result.

eminence:
I migrated my [procdump](https://github.com/eminence/procdump) project to using this PR, and it went pretty smoothly.  I had to make only a few changes:

* The `ProcessCgroup` struct got renamed to `ProcessCGroup`
* The `cgroups()` function now returns a `ProcResult<CGroupControllers>` instead of a `ProcResult<Vec<ProcessCgroup>>`
* Things like `stat.starttime()` now require an additional call to `with_system_info()`.  

I also see that some functions like `procfs::vmstat` are gone, replaced by `procfs::VmStat::current()`.  The introduction of the `VmStat` struct seems find, but for better API compatbility, it seems like we can still keep `procfs::vmstat`.

Since this PR is really quite large, my current inclination is to merge this, and then do additional iterations in additional merge requests.

afranchuk:
It sounds like you hit most/all of the purposeful breaking changes! Those changes were mainly to improve consistency in the API and naming (e.g. `ProcessCgroup` was renamed because there was `ProcessCgroup` and `CGroupController` previously with difference capitalization).

I think it's fine to keep some of those standalone functions. I actually thought I had left a TODO somewhere mentioning that, but I don't see it.

I considered making individual extension traits in `procfs` to avoid the need for `with_system_info()`, but I felt that might be somewhat cumbersome in the API (not to mention annoying to write).

afranchuk:
I ran a tool (`cargo-semver-checks`) to check compatibility. It seems like the tool has some bug with `pub use`, as it complained about a few enums and structs being removed when they actually are the same as they used to be (through a `pub use` of `procfs-core`). Otherwise with my latest commit there are no other valid complains aside from that renamed struct (though I guess that exposes the limitations of that tool since it didn't catch the method changes).

eminence:
Ok, let's get this merged, so other PRs can be rebased.  Many thanks @afranchuk for working on this

PR #259, opened 2023-03-17
Bitflags v2

tatref:
Update to bitflags v2, it's a big update

Changelog is not clean yet, so migration path is not complete
See https://github.com/bitflags/bitflags/issues/310

At the moment, I only updated the version, and implemented derived Copy, Clone, Debug, Hash, Eq, PartialEq, PartialOrd, Ord
From the mentioned PR, seems like we'll need to add some repr(transparent) if we want serde to work.

Maybe it's better to wait until the complete migration guide is finished

### DIFF
--- a/procfs-core/Cargo.toml
+++ b/procfs-core/Cargo.toml
@@ -14,11 +14,11 @@ rust-version.workspace = true
 
 [features]
 default = ["chrono"]
-serde1 = ["serde"]
+serde1 = ["serde", "bitflags/serde"]
 
 [dependencies]
 backtrace = { version = "0.3", optional = true }
-bitflags = "1.2"
+bitflags = { version = "2" }
 chrono = { version = "0.4.20", optional = true, features = ["clock"], default-features = false }
 hex = "0.4"
 serde = { version = "1.0", features = ["derive"], optional = true }
--- a/procfs-core/src/keyring.rs
+++ b/procfs-core/src/keyring.rs
@@ -10,6 +10,7 @@ use std::{collections::HashMap, io::BufRead, time::Duration};
 bitflags! {
     /// Various key flags
     #[cfg_attr(feature = "serde1", derive(Serialize, Deserialize))]
+    #[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, PartialOrd, Ord)]
     pub struct KeyFlags: u32 {
             /// The key has been instantiated
             const INSTANTIATED = 0x01;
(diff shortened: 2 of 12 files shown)


tatref:
I rebased from master, it should be good to go

procfs-core/Cargo.toml
```diff
@@ -18,7 +18,7 @@ serde1 = ["serde"]
 
 [dependencies]
 backtrace = { version = "0.3", optional = true }
-bitflags = "1.2"
+bitflags = { version = "2", features = ["serde"] }
```
eminence, line 21: This will unconditionally bring in `serde`, right?

To avoid that, I think we can have:

```toml
[dependencies]
bitflags = { version = "2" }

[features]
serde1 = ["serde", "bitflags/serde"]
```

(untested by me)

tatref:
Sorry i'm not super versed about dependencies/features selection

I tried the two following snippets:
```
[features]
serde1 = ["serde", "bitflags/serde"]

[dependencies]
bitflags = { version = "2" }
```

```
[features]
serde1 = ["serde", "bitflags/serde"]

[dependencies]
bitflags = { version = "2", no-default-features = true }
```

In both cases, `cargo tree` gives:

```
procfs v0.15.1 (/home/tatref/dev/linux-mem/procfs/procfs)
├── bitflags v2.3.1
│   └── serde v1.0.163
...
```

If I use `cargo tree --features serde1`, I get an additional `serde v1.0.163` as a direct dependency below `procfs v0.15.1`

Any idea?

eminence:
Did you make the change in both `procfs` and `procfs-core` ?

When I make this change locally, here's what I see:

```
❯ cargo tree 
procfs v0.15.1 (/nas/achin/devel/procfs/procfs)
├── bitflags v2.2.1
```

```
❯ cargo tree --features=serde1
procfs v0.15.1 (/nas/achin/devel/procfs/procfs)
├── bitflags v2.2.1
│   └── serde v1.0.152
│       └── serde_derive v1.0.152 (proc-macro)
│           ├── proc-macro2 v1.0.51
│           │   └── unicode-ident v1.0.6
│           ├── quote v1.0.23
│           │   └── proc-macro2 v1.0.51 (*)
│           └── syn v1.0.107
│               ├── proc-macro2 v1.0.51 (*)
│               ├── quote v1.0.23 (*)
│               └── unicode-ident v1.0.6
```

tatref:
Seems like this was fixed as part of  #270

PR #260, opened 2023-03-27
Don't hide process creation errors in all_processes

tatref:
This is followup of #188 

Current implementation hide `Process` creation errors in `all_processes`, making this function nearly useless in a resource constrained environment. Errors are silently ignored, making look like only a few processes exist.

The fact that FDs are created for each process is not necessarily an issue, but users of the library must have a way to know if any operation worked as expected or not.

This PR shouldn't change much for users, because the `Processes` are wrapped in a `Result` either way

Example usage

```rust

let all_processes: Vec<Process> = procfs::process::all_processes()
    .unwrap()
    .filter_map(|p| match p {
        Ok(p) => Some(p),                              // happy path
        Err(e) => match e {
            procfs::ProcError::NotFound(_) => None,    // process vanished during iteration
            procfs::ProcError::Io(e, path) => None,    // can match on path to decide if we can continue
            x => {
                panic!("Can't read process {x:?}");    // some unknown error
            }
        },
    })
    .collect();
```

### DIFF
--- a/procfs/src/process/mod.rs
+++ b/procfs/src/process/mod.rs
@@ -957,9 +957,7 @@ impl std::iter::Iterator for ProcessesIter {
             match self.inner.next() {
                 Some(Ok(entry)) => {
                     if let Ok(pid) = i32::from_str(&entry.file_name().to_string_lossy()) {
-                        if let Ok(proc) = Process::new_with_root(self.root.join(pid.to_string())) {
-                            break Some(Ok(proc));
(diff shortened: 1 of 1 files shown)


tatref:
I rebased from master

eminence:
Thanks, this change makes sense.  As a followup TODO, some additional documentation about strategies to handle errors might be good (similar to the comment you made in the PR)

PR #261, opened 2023-03-28
Add /proc/mounts

tatref:
`/proc/<pid>/mounts` is the same format, so we could reuse some code

### DIFF
--- a/procfs-core/src/mounts.rs
+++ b/procfs-core/src/mounts.rs
@@ -1,0 +1,107 @@
+use std::{collections::HashMap, io::BufRead};
+
+use super::ProcResult;
+use std::str::FromStr;
+
+#[cfg(feature = "serde1")]
+use serde::{Deserialize, Serialize};
+
+/// A mountpoint entry under `/proc/mounts`
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde1", derive(Serialize, Deserialize))]
+#[allow(non_snake_case)]
+pub struct MountEntry {
+    /// Device
+    pub fs_spec: String,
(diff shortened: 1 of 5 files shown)


eminence:
Thanks for all the PRs.  I think I want to get #257 merged first, as I'm sure it'll conflict with other things.

tatref:
I rebased from master.

A few changes compared to the first version:

* There is no `Mounts(pub Vec<MountEntry>)` anymore
* I used `impl Current for Vec<MountEntry>`, same for `FromBufRead`
* Simple usage is easy: `procfs::mounts().unwrap()`
* Advanced usage is more complicated: `Vec::<MountEntry>::from_buf_read(cursor).unwrap()`

What do you think?

procfs-core/src/mounts.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub fs_freq: u8,
+    /// Check
+    pub fs_passno: u8,
+}
+
+impl super::FromBufRead for Vec<MountEntry> {
+    fn from_buf_read<R: BufRead>(r: R) -> ProcResult<Self> {
+        let mut vec = Vec::new();
+
+        for line in r.lines().skip(1) {
```
eminence, line 32: Is this `skip(1)` right?

eminence:
Also can you update `support.md` ?

procfs-core/src/mounts.rs
tatref: I don't know what I did there... the file has no headers, so no reason to skip the first line

eminence:
Thanks

PR #264, opened 2023-05-04
Fixed two broken links to examples from the README

myz-dev:
Hi,
sorry if this is not the correct way to go about it, I just wanted to fix the URLs in the README.
I do not know much about pull requests :-(


eminence:
Looks great, thanks!

PR #266, opened 2023-05-09
Fix O_PATH for old kernels

### LINKED ISSUE #265: Process uid() fails on some kernels
Tested on 2.6.39 UEK kernel on Oracle Linux 6 x86_64

`Process::uid()` generates `Bad file descriptor`. This is due to `OFlags::PATH` beeing used to open the `/proc/<id>` directory:

https://github.com/eminence/procfs/blob/86e5ba008aa0eff720ce87b9fe4410e23ec4674d/procfs/src/process/mod.rs#L149-L152
https://github.com/eminence/procfs/blob/86e5ba008aa0eff720ce87b9fe4410e23ec4674d/procfs/src/process/mod.rs#L221-L224

This flag does not give access to some syscall (`fstat` in this case).

From the manpage (https://man7.org/linux/man-pages/man2/open.2.html):
```
O_PATH (since Linux 2.6.39)
[...]
              other file operations read, write, fchmod, fchown, fgetxattr, ioctl, mmap fail with the
              error EBADF.

              The following operations can be performed on the resulting
              file descriptor:

              *  [fstat(2)](https://man7.org/linux/man-pages/man2/fstat.2.html) (since Linux 3.6).
              *  [fstatfs(2)](https://man7.org/linux/man-pages/man2/fstatfs.2.html) (since Linux 3.12).
```

So for kernels >= 2.6.39 but < 3.6, we should not set the `OFlags::PATH` flag.
For kernels < 2.6.39, the flag seems to be ignored, so we can keep it.
For kernels >= 3.6, the flag gives better performance, so we should keep it

Can we add a simple `if kernel >= 3.6 { flags |= OFlags::PATH }`? Or is there a better way to do it?

I don't know if this should be a compile-time check, or runtime. I would personally prefer a runtime check because I reuse a single binary over multiple server, but that's just my personal use-case.

tatref:
Fixes #265

### DIFF
--- a/procfs/src/process/mod.rs
+++ b/procfs/src/process/mod.rs
@@ -147,8 +147,7 @@ impl FDInfo {
         // for 2.6.39 <= kernel < 3.6 fstat doesn't support O_PATH see https://github.com/eminence/procfs/issues/265
         let flags = match *crate::KERNEL {
             Ok(v) if v < KernelVersion::new(3, 6, 0) => OFlags::NOFOLLOW | OFlags::CLOEXEC,
-            Ok(v) if v >= KernelVersion::new(3, 6, 0) => OFlags::NOFOLLOW | OFlags::PATH | OFlags::CLOEXEC,
-            Ok(_) => unreachable!(),
+            Ok(_) => OFlags::NOFOLLOW | OFlags::PATH | OFlags::CLOEXEC,
             Err(_) => OFlags::NOFOLLOW | OFlags::PATH | OFlags::CLOEXEC,
         };
         let file = wrap_io_error!(root, rustix::fs::openat(dirfd, p, flags, Mode::empty()))?;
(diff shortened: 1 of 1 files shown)


procfs/src/process/mod.rs
```diff
@@ -144,15 +144,14 @@ impl FDInfo {
...
-                p,
-                OFlags::NOFOLLOW | OFlags::PATH | OFlags::CLOEXEC,
-                Mode::empty()
-            )
-        )?;
+        // for 2.6.39 <= kernel < 3.6 fstat doesn't support O_PATH see https://github.com/eminence/procfs/issues/265
+        let flags = match *crate::KERNEL {
+            Ok(v) if v < KernelVersion::new(3, 6, 0) => OFlags::NOFOLLOW | OFlags::CLOEXEC,
+            Ok(v) if v >= KernelVersion::new(3, 6, 0) => OFlags::NOFOLLOW | OFlags::PATH | OFlags::CLOEXEC,
+            Ok(_) => unreachable!(),
```
eminence, line 151: This is minor, but can we re-write the match to get rid of the unreachable statement?

```suggestion
            Ok(_) => OFlags::NOFOLLOW | OFlags::PATH | OFlags::CLOEXEC,
```

PR #269, opened 2023-05-12
Fix readme example

popsu:
Tried to run the example and it had a small error. procfs::page_size returns a u64, not a Result:
https://github.com/eminence/procfs/blob/a869cc7e3971dabf3da3d37d1702df62be392513/procfs/src/lib.rs#L341


eminence:
This is another thing I broke in #235

Thanks for the fix!

PR #275, opened 2023-06-29
Fix handling of false default-features.

smmalis37:
Without this it is impossible to turn off procfs-core's default-features.


eminence:
Nice, thank you

PR #281, opened 2023-08-01
Support /proc/net/snmp and /proc/net/snmp6

wfly1998:


eminence:
Just did a quick skim, and this looks great, thank you.  I am especially impressed with all of the nice docs.  Can you take a look at the CI failure, and also update `support.md` ?

wfly1998:
> Just did a quick skim, and this looks great, thank you. I am especially impressed with all of the nice docs. Can you take a look at the CI failure, and also update `support.md` ?

Sorry I overlooked the CI failure earlier.  An update has been pushed, please review it.

BTW, most of the docs are from the official documentation, e.g., [RFC1213](https://datatracker.ietf.org/doc/html/rfc1213) and [SNMP counter](https://docs.kernel.org/networking/snmp_counter.html).  I just copied them, but some non-rfc fields do not have any doc 😢

PR #283, opened 2023-08-16
Don't convert meminfo units.

### LINKED ISSUE #282: Consider leaving meminfo values as kilobytes/kibibytes
Currently all values reported in the Meminfo type are converted to bytes from their reported value in kibibytes. This can be confusing for users who expect the values reported to be unchanged from the values shown in /proc/meminfo. Given that 0.16.0 is a breaking change release coming soon, could this behavior be changed? Would you be open to a PR making this change?

smalis-msft:
Fixes #282

### DIFF
--- a/procfs-core/src/meminfo.rs
+++ b/procfs-core/src/meminfo.rs
@@ -3,16 +3,6 @@ use super::{expect, from_str, ProcResult};
 use serde::{Deserialize, Serialize};
 use std::{collections::HashMap, io};
 
-fn convert_to_kibibytes(num: u64, unit: &str) -> ProcResult<u64> {
-    match unit {
-        "B" => Ok(num),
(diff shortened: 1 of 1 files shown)


procfs-core/src/meminfo.rs
```diff
@@ -320,11 +311,9 @@ impl super::FromBufRead for Meminfo {
 
             let value = from_str!(u64, value);
 
-            let value = if let Some(unit) = unit {
-                convert_to_kibibytes(value, unit)?
-            } else {
-                value
-            };
+            if let Some(unit) = unit {
+                assert_eq!(unit, "kB", "/proc/meminfo field wasn't in kB.");
```
tatref, line 315: You should probably return an Err instead of panicking

eminence:
Closing as discussed in https://github.com/eminence/procfs/issues/282

PR #292, opened 2023-11-05
cgroups: skip empty controllers

eliad-wiz:
In case of cgroupv2, the controllers list is empty.

Splitting it by comma results in a vector with a single empty element ([""]), instead of an empty vector ([]).


eminence:
LGTM, thanks

PR #294, opened 2023-11-27
Correct copy-pasted Process::task_from_tid() doc

ryoqun:
the doc comment of `Process::task_from_tid()` is wrongly copy-pasted from `Process::task_main_thread()`.


eminence:
Nice catch, thank you!

PR #296, opened 2023-12-06
Implemented proc/crypto parsing 

Hwatwasthat:
Hi, I have implemented parsing for the /proc/crypto file, as I noticed it wasn't implemented in the support file. I have tried to add in types where it felt relevant so it can restrict the output slightly and make handling the output a bit easier than throwing strings around. I've put in tests for various possibilities but admittedly they're far from exhaustive! I've tested on my local machine and a raspberry pi but nothing further.

### DIFF
--- a/procfs-core/src/crypto.rs
+++ b/procfs-core/src/crypto.rs
@@ -1,0 +1,532 @@
+use crate::{expect, FromBufRead, ProcError, ProcResult};
+
+#[cfg(feature = "serde1")]
+use serde::{Deserialize, Serialize};
+use std::{
+    collections::HashMap,
+    convert::TryFrom,
+    io::BufRead,
+    iter::{once, Peekable},
+    str::FromStr,
+};
+
+/// Represents the data from `/proc/crypto`.
+///
+/// Each block represents a cryptographic implementation that has been registered with the kernel.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde1", derive(Serialize, Deserialize))]
+pub struct CryptoTable {
+    pub crypto_blocks: HashMap<String, CryptoBlock>,
+}
+
+/// Format of a crypto implementation represented in /proc/crypto.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde1", derive(Serialize, Deserialize))]
+pub struct CryptoBlock {
+    pub name: String,
+    pub driver: String,
+    pub module: String,
+    pub priority: isize,
+    pub ref_count: isize,
+    pub self_test: SelfTest,
+    pub internal: bool,
+    pub fips_enabled: bool,
+    pub crypto_type: Type,
+}
+
+impl FromBufRead for CryptoTable {
+    fn from_buf_read<R: BufRead>(r: R) -> ProcResult<Self> {
+        let mut lines = r.lines().flatten().peekable();
+        let mut crypto_blocks = HashMap::new();
+        loop {
+            let line = match lines.next() {
+                Some(line) => line,
+                // We got to the end of the file
+                None => break,
+            };
+            // Just skip empty lines
+            if !line.is_empty() {
+                let mut split = line.split(":");
+                let name = expect!(split.next());
+                if name.trim() == "name" {
+                    let name = expect!(split.next()).to_string();
+                    let block = CryptoBlock::from_iter(&mut lines, name.as_str())?;
+                    crypto_blocks.insert(name, block);
+                }
+            }
+        }
+
+        Ok(CryptoTable { crypto_blocks })
+    }
+}
+
+impl CryptoBlock {
+    fn from_iter<T: Iterator<Item = String>>(iter: &mut Peekable<T>, name: &str) -> ProcResult<Self> {
+        let driver = parse_line(iter, "driver", &name)?;
+        let module = parse_line(iter, "module", &name)?;
+        let priority = from_str!(isize, &parse_line(iter, "priority", &name)?);
+        let ref_count = from_str!(isize, &parse_line(iter, "refcnt", &name)?);
+        let self_test = SelfTest::try_from(parse_line(iter, "selftest", &name)?.as_str())?;
+        let internal = parse_bool(iter, "internal", name)?;
+        let fips_enabled = parse_fips(iter, name)?;
+        let crypto_type = Type::from_iter(iter, name)?;
+        Ok(CryptoBlock {
+            name: name.to_string(),
+            driver,
+            module,
+            priority,
+            ref_count,
+            self_test,
+            internal,
+            fips_enabled,
+            crypto_type,
+        })
+    }
+}
+
+/// Potential results for selftest.
+#[derive(Debug, Clone, PartialEq)]
+#[cfg_attr(feature = "serde1", derive(Serialize, Deserialize))]
+pub enum SelfTest {
+    Passed,
+    Unknown,
+}
+
+impl TryFrom<&str> for SelfTest {
+    type Error = ProcError;
+
+    fn try_from(value: &str) -> Result<Self, ProcError> {
+        Ok(match value {
+            "passed" => Self::Passed,
+            "unknown" => Self::Unknown,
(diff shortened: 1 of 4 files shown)


eminence (changes requested):
Hi, thanks for working on this!  This is a good start

I made a few in-line comments.  A few additional general comments:

* Can you add a test that will actually parse a live `/proc/crypto` file?
* I don't see an obvious way to construct a `CryptoTable`
* There's a few very minor things that `cargo clippy` will report

procfs-core/src/crypto.rs
```diff
@@ -0,0 +1,541 @@
...
+#[cfg_attr(feature = "serde1", derive(Serialize, Deserialize))]
+pub struct Skcipher {
+    r#async: bool,
+    block_size: usize,
+    min_key_size: usize,
+    max_key_size: usize,
+    iv_size: usize,
+    chunk_size: usize,
+    walk_size: usize,
+}
```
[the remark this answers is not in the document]
Hwatwasthat: I could rename to `async_capable`, as that's what is essentially saying. You're right about the fields, not sure why I did that.

procfs-core/src/crypto.rs
```diff
@@ -0,0 +1,541 @@
...
+                if name.trim() == "name" {
+                    let name = expect!(split.next()).trim().to_string();
+                    let block = CryptoBlock::from_iter(&mut lines, name.as_str())?;
+                    crypto_blocks.insert(name, block);
+                }
+            }
+        }
+
+        Ok(CryptoTable { crypto_blocks })
+    }
```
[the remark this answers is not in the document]
Hwatwasthat: The reason I've done the loop like this is so I can carry on using the iterator, saves a bit on memory and lets the iterator keep track of where I am in the file. I could gather everything in a block into a temporary vec, roughly:
```
let mut buf = Vec::new();
for line in lines {
    if !line.empty() {
        buf.push(line);
    } else {
        // Would need to update logic here
        let block = CryptoBlock::from_iter(&mut buf)?:
       crypto_blocks.insert(block);
    }
}
```

Using the so far stable assumption that its crypto blocks separated by lines and ended by an EOF, tracks with the code that I've found so far.

I've got a work in progress that doesn't use flatten, it just muddies up the function signatures a bit but isn't that devastating.

procfs-core/src/crypto.rs
```diff
@@ -0,0 +1,541 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq)]
+#[cfg_attr(feature = "serde1", derive(Serialize, Deserialize))]
+pub struct Unknown {
+    fields: HashMap<String, String>,
+}
+
+impl Unknown {
+    fn parse<T: Iterator<Item = String>>(iter: &mut T, unknown_name: &str, _name: &str) -> ProcResult<Self> {
```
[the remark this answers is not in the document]
Hwatwasthat: That's fair, will update that.

eminence:
> For the test, you mean as in load a local file or just paste in the contents of an existing file?

I mean open `/proc/crypto` and make sure it can be parsed.  This is relatively common in this crate, and it's useful because I can clone this repo to any computer and just run `cargo test` to make sure that everything can be parsed.

> At the moment there is only a way to create one from an iterator

Sorry, I didn't quite follow this.  Can you give an example about how one would create a `CryptoTable`?    I think probably you can `impl Current for CrypotoTable` which means that you'd be able to run `CryptoTable::current()` to get an instance of `CryptoTable`

Hwatwasthat:
I can implement those, I just normally try and keep it all in test and not touch the system. I can implement current for a table, currently it's only implemented that if you pass an iterator of bytes that contains a crypto table it will parse from that. I'll get a push to you this weekend that will implement these.

Hwatwasthat:
I think I've got everything, if you could give it a look over :) There are some clippy lints outside of the crypto sections that I have left untouched, as I don't want to mess with other folks code!

eminence:
I'm looking at the `/proc/crypto` file on my computer, and I see several blocks with the same name.  For example:

```
name         : stdrng
driver       : drbg_nopr_hmac_sha512
module       : drbg
priority     : 221
refcnt       : 2
selftest     : passed
internal     : no
type         : rng
seedsize     : 0

name         : stdrng
driver       : drbg_nopr_hmac_sha256
module       : drbg
priority     : 220
refcnt       : 1
selftest     : passed
internal     : no
type         : rng
seedsize     : 0
```

Since the crypto table is a HashMap, only the last `stdrng` gets saved.  Is that what you intended with this code?

procfs-core/src/crypto.rs
```diff
@@ -0,0 +1,570 @@
...
+module       : kernel
+priority     : 0
+refcnt       : 2
+selftest     : passed
+internal     : no
+type         : compression"#;
+        let mut iter = block.lines().map(|s| Ok(s.to_string())).peekable();
+        let block = CryptoBlock::from_iter(&mut iter, "deflate");
+        assert!(block.is_ok());
+        let block = block.unwrap();
```
eminence, line 448: The assertion on `block.is_ok()` is redundant, because `block` gets unwrapped on the next line, so you can just remove this line.

```suggestion
        let block = block.unwrap();
```

procfs-core/src/crypto.rs
```diff
@@ -0,0 +1,570 @@
...
+                }
+            }
+        }
+
+        Ok(CryptoTable { crypto_blocks })
+    }
+}
+
+impl CryptoTable {
+    pub fn get<T: AsRef<str>>(&self, target: &T) -> Option<&CryptoBlock> {
```
eminence, line 60:
```suggestion
    pub fn get<T: AsRef<str>>(&self, target: T) -> Option<&CryptoBlock> {
```

procfs-core/src/crypto.rs
Hwatwasthat: Hey, I've modified to be an expect, just so we can have an error message in there to make it easier to understand the context when the test fails.

Hwatwasthat:
That was not the intention. My reckoning is that the easiest way to resolve this is to change from a `Hashmap<String, CryptoBlock` to a `Hashmap<String, Vec<CryptoBlock>>`. This removes the loss of duplicates but maintains the ease of the interface, having to know the exact name of a driver is probably not the most desired outcome when using this interface, and it still allows for more fine grained searching later, without the overhead of a double hashmap (I doubt there are that many duplicates).

Hwatwasthat:
Added in a version that has a `Hashmap<String, Vec<CryptoBlock>>`. This adds some extra overhead to the type sadly but probably won't be that hurtful. Duplicates are just pushed to the vec when it happens, and that can be further broken down by the caller.

eminence:
Looks pretty good on a quick skim, will do a more detailed review later.  Thanks!

Hwatwasthat:
Hey, just checking in if you've had a chance to look and see if anything else needs an update?

eminence:
Thanks for the ping, I'll take a look Very Soon Now

procfs-core/src/crypto.rs
```diff
@@ -0,0 +1,597 @@
...
+        })
+    }
+}
+
+/// Enumeration of potential types and their associated data. Unknown at end to catch unrecognised types.
+#[derive(Debug, Clone, PartialEq)]
+#[cfg_attr(feature = "serde1", derive(Serialize, Deserialize))]
+pub enum Type {
+    /// Symmetric Key Cipher
+    Skcipher(Skcipher),
```
eminence, line 122: I think I might prefer using a struct variant, like:

```rust
pub enum Type {
    /// Symmetric Key Cipher
    Skcipher{async_capable: bool, block_size: bool, min_key_size: bool, ...}
     ...
}
```

It makes the whole enum a bit more complicated, but also you have less types overall (because you can get rid of the `Skcipher` struct).

But this is a minor point, no need to change this now

eminence (approved):

eminence:
Thank you for the contribution, and your patience with the long review!

PR #298, opened 2024-01-10
Add oom_score_adj Process methods

futpib:
I took note of this comment: https://github.com/eminence/procfs/pull/139#pullrequestreview-737802630:

> So the way I read this is: while the final OOM "score" will be somewhere between 0 and 1000, the value reported in oom_score_adj can be between -1000 and 1000. So I think the oom_score_adj function should return a i16 not u32.

And changed `oom_score` type to u16 too, since it too only ranges from 0 to 1000.

Previous attempt: https://github.com/eminence/procfs/pull/139


eminence:
LGTM, thanks!

PR #300, opened 2024-01-20
Add MountInfos::iter

rusty-snake:
This is a convenience for `mountinfo.0.iter()`.

### DIFF
--- a/procfs-core/src/process/mount.rs
+++ b/procfs-core/src/process/mount.rs
@@ -49,6 +49,13 @@ bitflags! {
 /// This data is taken from the `/proc/[pid]/mountinfo` file.
 pub struct MountInfos(pub Vec<MountInfo>);
 
+impl MountInfos {
+    /// Returns an borrowed iterator.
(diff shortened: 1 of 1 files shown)


procfs-core/src/process/mount.rs
```diff
@@ -49,6 +49,13 @@ bitflags! {
 /// This data is taken from the `/proc/[pid]/mountinfo` file.
 pub struct MountInfos(pub Vec<MountInfo>);
 
+impl MountInfos {
+    /// Returns an borrowed iterator.
+    pub fn iter(&self) -> std::slice::Iter<'_, MountInfo> {
```
eminence, line 54: This is a minor point, but I wonder if we could do:

```suggestion
    pub fn iter(&self) -> impl Iterator<Item = &MountInfo> {
```

to hide the `std::slice::Iter` concrete type.   In practice this doesn't matter much because the inner `Vec` is part of the public API of `MountInfos` (and so `std::slice::Iter` is kind of part of this API too)

eminence:
LGTM, thanks!

PR #303, opened 2024-02-11
feat: move PhysicalPageFlags to procfs-core

tatref:
Hi,

I needed to use `PhysicalPageFlags` for a Windows project, so I moved it to procfs-core

I made `parse_info` pub

I removed the import of `CGroupController` to remove a warning

Maybe we could move `KPageFlags` also, but we would need to also move `FileWrapper` that is part of procfs.
What do you think?

### DIFF
--- a/procfs-core/src/kpageflags.rs
+++ b/procfs-core/src/kpageflags.rs
@@ -1,0 +1,90 @@
+use bitflags::bitflags;
+
+#[cfg(feature = "serde1")]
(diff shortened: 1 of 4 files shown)


eminence:
Regarding `KPageFlags`, we could maybe move the `from_custom_root` into some trait that only exists in procfs (not procfs-core).

I need to review how procfs types are constructed in general, but I think we aren't very consistent.  For example, I think several types should `impl Current` but don't at the moment

Issue #295, opened 2023-11-30
Add support for parsing /proc/net/netstat


chengshuyi:
/proc/net/netstat is a very important file and is often used to diagnose problems related to network packet loss. Therefore, many monitoring tools will check this file regularly, so it is valuable to add analysis to this file.

mwinters0:
FYI I'm WIP on this with a different approach from the previous incomplete PR. This approach will be compatible with both past and future kernels.

PR #304, opened 2024-02-14
feat: add initial implementation for parsing /proc/net/netstat

etolbakov:
### Description
added parsing for `/proc/net/netstat`

### How was this PR tested?
added a unit test

### Related issues
https://github.com/eminence/procfs/issues/295

### DIFF
--- a/procfs-core/src/netstat.rs
+++ b/procfs-core/src/netstat.rs
@@ -1,0 +1,625 @@
+use crate::ProcResult;
+#[cfg(feature = "serde1")]
+use serde::{Deserialize, Serialize};
+use std::io::BufRead;
+
+/// Represents the data from `/proc/net/netstat`.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde1", derive(Serialize, Deserialize))]
+pub struct Netstat {
+    // TcpExt
+    /// The number of SYN cookies sent
+    pub tcp_ext_syncookies_sent: u64,
+    /// The number of SYN cookies received
+    pub tcp_ext_syncookies_recv: u64,
+    // The number of invalid SYN cookies received
+    pub tcp_ext_syncookies_failed: u64,
+    /// The number of resets received for embryonic SYN_RECV sockets
+    pub tcp_ext_embryonic_rsts: u64,
+    /// The number of packets pruned from receive queue because of socket buffer overrun
+    pub tcp_ext_prune_called: u64,
+    /// The number of packets pruned from receive queue
+    pub tcp_ext_rcv_pruned: u64,
+    /// The number of packets dropped from out-of-order queue because of socket buffer overrun
+    pub tcp_ext_ofo_pruned: u64,
+    /// The number of ICMP packets dropped because they were out-of-window
+    pub tcp_ext_out_of_window_icmps: u64,
+    /// The number of ICMP packets dropped because socket was locked
+    pub tcp_ext_lock_dropped_icmps: u64,
+    // The number of arp packets filtered
+    pub tcp_ext_arp_filter: u64,
+    /// The number of TCP sockets finished time wait in fast timer
+    pub tcp_ext_tw: u64,
+    /// The number of time wait sockets recycled by timestamp
+    pub tcp_ext_tw_recycled: u64,
+    /// The number of TCP sockets finished time wait in slow timer
+    pub tcp_ext_tw_killed: u64,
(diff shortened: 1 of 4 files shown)


etolbakov:
@eminence 
Could you please take a look at the change when you have time? Looking forward getting feedback.

1) I've noticed [the comment](https://github.com/eminence/procfs/blob/e734c74ee2d60a216bc87c1b0930a61448e8c82c/procfs-core/src/net.rs#L7) in `net.rs` so maybe the code should be moved there.  I decided to come up with a dedicated file to make a review process simpler.
2) some of the comments still have TODO, I haven't found any meaningful description for them.

procfs-core/src/netstat.rs
```diff
@@ -0,0 +1,625 @@
+use crate::ProcResult;
+#[cfg(feature = "serde1")]
+use serde::{Deserialize, Serialize};
+use std::io::BufRead;
+
+/// Represents the data from `/proc/net/netstat`.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde1", derive(Serialize, Deserialize))]
+pub struct Netstat {
```
eminence, line 9: Do you think we should mark this struct as `#[non_exhaustive]`, in case new fields get added in newer kernels?

procfs-core/src/netstat.rs
```diff
@@ -0,0 +1,625 @@
...
+            tcp_ext_tcp_rcv_qdrop: from_str!(u64, expect!(tcp_ext.next())),
+            tcp_ext_tcp_wqueue_too_big: from_str!(u64, expect!(tcp_ext.next())),
+            tcp_ext_tcp_fast_open_passive_alt_key: from_str!(u64, expect!(tcp_ext.next())),
+            tcp_ext_tcp_timeout_rehash: from_str!(u64, expect!(tcp_ext.next())),
+            tcp_ext_tcp_duplicate_data_rehash: from_str!(u64, expect!(tcp_ext.next())),
+            tcp_ext_tcp_ds_ack_recv_segs: from_str!(u64, expect!(tcp_ext.next())),
+            tcp_ext_tcp_ds_ack_ignored_dubious: from_str!(u64, expect!(tcp_ext.next())),
+            tcp_ext_tcp_migrate_req_success: from_str!(u64, expect!(tcp_ext.next())),
+            tcp_ext_tcp_migrate_req_failure: from_str!(u64, expect!(tcp_ext.next())),
+            tcp_ext_tcp_ecn_rehash: from_str!(u64, expect!(tcp_ext.next())),
```
eminence, line 534: My 6.1.0 kernel doesn't have this field:

```
---- net::tests::test_netstats stdout ----
thread 'net::tests::test_netstats' panicked at procfs/src/net.rs:264:35:
called `Result::unwrap()` on an `Err` value: InternalError(bug at procfs-core/src/netstat.rs:534 (please report this procfs bug)
Internal Unwrap Error: NoneError)
```

We're going to want to find some older kernels to test this on.  I'll try to capture some examples of `/proc/net/netstat` on some random old machines

procfs-core/src/netstat.rs
etolbakov: yeah, good idea! I'll add `#[non_exhaustive]`

procfs-core/src/netstat.rs
etolbakov: Thank you! I'll provide the info about the kernel version I took it from

etolbakov:
Hi Andrew @eminence 
Thanks for taking a look at the PR! I'll address the suggestions shortly

procfs-core/src/netstat.rs
etolbakov: @eminence 
I've confirmed that these parameters 
```
TcpExt:....
TcpDuplicateDataRehash TCPDSACKRecvSegs TCPDSACKIgnoredDubious TCPMigrateReqSuccess TCPMigrateReqFailure TCPECNRehash
```
come from 
```
sh-4.2$ uname -r
5.10.205-195.807.amzn2.x86_64
```
Not sure if that's too representative/widely used

etolbakov:
Hello Andrew @eminence !
Could you please let me know if anything else needed in the scope of this PR?

eminence:
We still have to fix the issue about fields that are only in newer kernels.  These should become `Option` fields in `Netstat` struct

etolbakov:
Thanks understood, will adjust! 
I wonder if you had time to get outputs from older kernels for tests?

PR #306, opened 2024-03-01
Fix documentation for `Meminfo.s_reclaimable`.

carletes:
It incorrecly states that this memory cannot be reclaimed (that would be the field `s_unreclaim`).


eminence:
Nice catch, thank you

PR #308, opened 2024-04-30
Remove lazy-static dependency

notgull:
lazy-static is on its way to being deprecated. I'm trying to remove it
from my dependency tree. Therefore I've removed lazy-static from this
crate in this commit.

Modifications made:

- PAGESIZE and TICKS_PER_SECOND appear to be unused so I've removed
  them.
- KERNEL can be replaced with racy atomic initialization.
- RANDOM_ROOT doesn't really need to be allocated on the heap, so I've
  replaced it with manual Path::new.

### DIFF
--- a/procfs/Cargo.toml
+++ b/procfs/Cargo.toml
@@ -20,7 +20,6 @@ serde1 = ["serde", "procfs-core/serde1"]
 procfs-core = { path = "../procfs-core", version = "0.16.0-RC1", default-features = false }
 rustix = { version = "0.38.19", features = ["fs", "process", "param", "system", "thread"] }
 bitflags = { version = "2.0", default-features = false }
-lazy_static = "1.0.2"
 chrono = {version = "0.4.20", optional = true, features = ["clock"], default-features = false }
 hex = "0.4"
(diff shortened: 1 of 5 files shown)


eminence:
Thanks, on quick skim this looks good, will review more carefully later

notgull:
@eminence Have you had a chance to review this more carefully?

eminence:
LGTM, thanks!

eminence:
> I'm trying to remove it from my dependency tree.

I'm always curious to know where/how `procfs` is used, so if you're comfortable sharing, where is `procfs` being used in your deptree?

notgull:
> I'm always curious to know where/how `procfs` is used, so if you're comfortable sharing, where is `procfs` being used in your deptree?

I was taking [`bpftop`](https://github.com/Netflix/bpftop) for a spin when I realized that `procfs` was pulling in `lazy-static` in its dependency tree.

notgull:
@eminence Are you planning on making a release with this change?

eminence:
Let me try to get a release out in a day or so

eminence:
This is now published at 0.17.0

PR #311, opened 2024-06-01
Add support for /dev/devices

yamaura:
Thank you for this amazing library! I found it incredibly useful. I implemented parser of /proc/devices and test.

Notes:

- The type for `major` has been same with the Linux kernel code.
- While `BlockDeviceEntry` can be disabled via `CONFIG_BLOCK`, but it has not been made an `Option`.

### DIFF
--- a/procfs-core/src/devices.rs
+++ b/procfs-core/src/devices.rs
@@ -9,7 +9,6 @@ use serde::{Deserialize, Serialize};
 /// Device entries under `/proc/devices`
 #[derive(Debug, Clone)]
 #[cfg_attr(feature = "serde1", derive(Serialize, Deserialize))]
-#[allow(non_snake_case)]
 pub struct Devices {
     /// Character devices
     pub char_devices: Vec<CharDeviceEntry>,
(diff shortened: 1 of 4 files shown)


eminence:
Thank you!  I'm sorry for taking so long to notice and review this PR.

This looks good, only a few comments:

> While `BlockDeviceEntry` can be disabled via `CONFIG_BLOCK`, but it has not been made an `Option`.

The API you've added here looks good, but can you update the doc comment to say that the `block_devices` list can be empty if the kernel doesn't support block devices?


Also, these structs have `#[allow(non_snake_case)]` but that doesn't seem to be needed.

eminence:
Thanks!  Quick question:  why is `CharDeviceEntry::major` a `u32` while `BlockDeviceEntry::major` an `i32` ?

eminence:
Thanks for those links.  My general policy is to prioritize alignment with the kernel (though I think I deviate from that in a few spots).  A comment in the code with these links would be useful to anyone reading this code in the future, but that can happen later.

Thanks for the pull request!

Issue #290, opened 2023-10-29
Improve our examples


eminence:
We have several examples in the `procfs/examples` folder.  Some of them could use additional love and attention.

For example, the [`partitions.rs`](https://github.com/eminence/procfs/blob/71addc0766626dca44f5be3d771b204f6ed1fc3d/procfs/examples/partitions.rs) example doesn't do anything besides print out the `Debug` impl.

I've love if some of these examples were enhanced to be a little more polished.  

This might mean printing fields in a nice table (like [`interface_stats.rs`](https://github.com/eminence/procfs/blob/71addc0766626dca44f5be3d771b204f6ed1fc3d/procfs/examples/interface_stats.rs) or [`lslocks.rs`](https://github.com/eminence/procfs/blob/71addc0766626dca44f5be3d771b204f6ed1fc3d/procfs/examples/lslocks.rs)), or doing something a little more involved (like `lslocks.rs`) that shows a semi-realworld example of how you might use these functions.

Note that there is a balancing act here -- we don't want these examples to be *too* complicated, because they ought to be short, readable, and primarily about one feature in particular.

PR #314, opened 2024-07-25
Updated some examples to a bit more than printing debug.

Hwatwasthat:
I just updated some of the existing examples to have a slightly better visual output. I haven't pushed it any further than just making them a bit nicer to look at.


eminence:
CC #290

Thanks!

PR #315, opened 2024-08-05
Implemented crypto example

Hwatwasthat:
Produced an example program for running the /dev/crypto parser. Added a small description to the example readme file.


eminence:
Always happy to have more examples.  Thanks!

PR #320, opened 2024-11-11
Enable `generate-link-to-definition` rustdoc feature on docs.rs

GuillaumeGomez:
This feature is very useful when browsing source code directly on docs.rs (from the source link). Example with `syn` crate: https://docs.rs/syn/latest/src/syn/lib.rs.html#950


eminence:
Very cool, thanks!

PR #327, opened 2024-12-27
replace hardcoded crate name by env var that returns the crate's name dynamically

PauloMigAlmeida:
Disclaimer: That's a trivial patch and it doesn't add any new functionality.


eminence:
I'm not sure why we need this, but why not.

PR #334, opened 2025-02-05
Parse /proc/pid/status "Groups" field as u32

Jongy:
`Groups` is printed as unsigned:
https://github.com/torvalds/linux/blob/5c8c229261f14159b54b9a32f12e5fa89d88b905/fs/proc/array.c#L199-L203

I got an error like this:
```
Err(InternalError(bug at procfs-core/src/process/status.rs:309 (please report this procfs bug)
Internal Unwrap Error: Failed to parse i ("3555555555") as a T: number too large to fit in target type))
```
(procfs/procfs-core 0.17.0)

I suspect it's the `Groups` , other users of `parse_list` are `NStgid` `NSpid` `NSpgid` `NSsid` (which are also printed unsigned...) but the kernel limits them to 1 billion per the docs here: https://github.com/torvalds/linux/blob/5c8c229261f14159b54b9a32f12e5fa89d88b905/include/linux/threads.h#L30-L35 so it's safe to assume the issue I observed is from `Groups`, and we don't need to change their type (`pid_t` is truly signed).

Here's is a small reproducer program:
```
#include <unistd.h>
#include <stdlib.h>
#include <grp.h>
#include <stdio.h>
#include <errno.h>
#include <string.h>

int main() {
    gid_t new_gid = 3555555555;
    int ngroups;
    gid_t *groups;

    // Get current number of supplementary groups
    ngroups = getgroups(0, NULL);
    if (ngroups == -1) {
        printf("getgroups failed: %s\n", strerror(errno));
        return 1;
    }

    // Allocate space for current groups plus one more
    groups = malloc((ngroups + 1) * sizeof(gid_t));
    if (!groups) {
        printf("malloc failed\n");
        return 1;
    }

    // Get current supplementary groups
    if (getgroups(ngroups, groups) == -1) {
        printf("getgroups failed: %s\n", strerror(errno));
        free(groups);
        return 1;
    }

    // Add new group
    groups[ngroups] = new_gid;
    ngroups++;

    // Set new supplementary groups list
    if (setgroups(ngroups, groups) == -1) {
        printf("setgroups failed: %s\n", strerror(errno));
        free(groups);
        return 1;
    }

    free(groups);
    printf("Successfully added GID %u\n", new_gid);

    sleep(9999999);
    return 0;
}
```
`proc.status()` on this PID fails on master and works on this PR.

### DIFF
--- a/procfs-core/src/lib.rs
+++ b/procfs-core/src/lib.rs
@@ -404,6 +404,13 @@ impl FromStrRadix for u64 {
         u64::from_str_radix(s, radix)
     }
 }
+
+impl FromStrRadix for u32 {
+    fn from_str_radix(s: &str, radix: u32) -> Result<u32, std::num::ParseIntError> {
+        u32::from_str_radix(s, radix)
+    }
+}
+
 impl FromStrRadix for i32 {
     fn from_str_radix(s: &str, radix: u32) -> Result<i32, std::num::ParseIntError> {
         i32::from_str_radix(s, radix)
--- a/procfs-core/src/process/status.rs
+++ b/procfs-core/src/process/status.rs
@@ -53,7 +53,7 @@ pub struct Status {
     /// Number of file descriptor slots currently allocated.
     pub fdsize: u32,
     /// Supplementary group list.
-    pub groups: Vec<i32>,
+    pub groups: Vec<u32>,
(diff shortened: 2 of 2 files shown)


eminence:
Nice find, thank you!

PR #335, opened 2025-02-05
Make get_pressure and parse_pressure_record public

SebastiaanYN:
When reading CPU pressure from the cgroup fs the `full` line is included. However, in the [CpuPressure](https://docs.rs/procfs/latest/procfs/struct.CpuPressure.html) struct only `some` is provided. Making the parser functions public allows reading the `full` line manually.

Alternatively, if preferred, the CpuPressure struct could be changed to include the `full` line when present.

### DIFF
--- a/procfs-core/src/pressure.rs
+++ b/procfs-core/src/pressure.rs
@@ -104,7 +104,14 @@ fn get_total(map: &HashMap<&str, &str>) -> ProcResult<u64> {
     )
 }
 
-fn parse_pressure_record(line: &str) -> ProcResult<PressureRecord> {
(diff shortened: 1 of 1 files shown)


eminence:
These changes look fine to me.

If I'm reading things correctly, `/proc/pressure/cpu` will always contain the `full` line, but the data will only be meaningful when we're in a cgroup.   In addition to this PR here (which can make is possible to easily parse files like `/sys/fs/cgroup/user.slice/cpu.pressure` so :thumbsup: for this), I'd also be in favor of adding the full `PressureRecord` to `CpuPressure`

PR #336, opened 2025-02-07
Use AsRef<Path> for user-facing path parameters

Jongy:
Allow users to pass `Path` etc. `FileWrapper::open_at` already supports that.


eminence:
LGTM

Jongy:
Thanks :)

PR #337, opened 2025-02-16
Check process name in test_proc_status_for_kthreadd

NoisyCoil:
Hi again. While completing the update of procfs in Debian, Fabian Grünbichler discovered a bug in `test_proc_status_for_kthreadd`: some containers (or namespaces) may have a process with PID 2 which isn't kthreadd, which will make the test fail. This MR adds an explicit check for the process' name. I was given permission from Fabian to use his fix for this PR.

Cheers!


eminence:
LGTM, thanks!

PR #339, opened 2025-03-01
Derive PartialEq & Eq for FDTarget

Jongy:
Comparing FDTargets for equality is useful.


eminence:
seems reasonable to me, thanks!

PR #341, opened 2025-03-08
Add Process.open_relative_flags

Jongy:
This allows to open files from `/proc/pid` in other modes, such as `O_WRONLY` or `O_PATH`.


eminence:
I'm not quite sure when you'd need to use this, but this seems reasonable

PR #342, opened 2025-03-10
Fix README.md example compilation

rantala:
`procfs::ticks_per_second()` is infallible, update README.md accordingly.


eminence:
Nice find, thank you!

PR #345, opened 2025-03-11
Fix for trait derivation on wasm targets

fzylogic:
The Namespaces struct uses OsStr, which is only available on Windows and Unix systems, leaving wasm targets unable to use the serde1 feature to deserialize other struct data. This change skips derivation of Serialize and Deserialize when targeting wasm as the build environment.


eminence:
Looks reasonable to me, thanks!

PR #348, opened 2025-05-14
Add /proc/buddyinfo parsing

jwake:
Added parsing code for `/proc/buddyinfo`, and unit tests for the parser.

Also added a `Pages` type over `u64` to make it more obvious that the data in `/proc/buddyinfo/` is measured in terms of memory pages and not bytes, kilobytes or anything else, rather than making the parsing code also query the system page size. I'm a little torn about whether or not to add conversions eg. from `Pages` to `usize` that factor in the current system page size, or leave it up to the API consumer.

### DIFF
--- a/procfs-core/src/types.rs
+++ b/procfs-core/src/types.rs
@@ -1,0 +1,79 @@
+macro_rules! gen_ops {
+    (@gen_op $type:ident, $underlying:ident, $opTrait:ident, $opFn: ident, $op:tt) =>
+    {
+        impl<T: Into<$underlying>> std::ops::$opTrait<T> for $type {
+            type Output = $type;
+
+            fn $opFn(self, rhs: T) -> Self::Output {
+                $type(self.0 $op rhs.into())
+            }
+        }
+    };
+    (@gen_op_assign $type:ident, $underlying:ident, $opTrait:ident, $opFn: ident, $op: tt) =>
+    {
+        impl<T: Into<$underlying>> std::ops::$opTrait<T> for $type {
+            fn $opFn(&mut self, rhs: T) {
+                self.0 $op rhs.into()
+            }
+        }
+    };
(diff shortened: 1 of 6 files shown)


eminence:
Sorry for taking so long to review this.  It looks good to me.  I don't understand the CI failures, and they happened so long ago that I can't retry them.  

> I'm a little torn about whether or not to add conversions eg. from Pages to usize that factor in the current system page size, or leave it up to the API consumer.

There is some precedent for this, for example in the [rss_bytes](https://docs.rs/procfs/latest/procfs/process/struct.Stat.html#method.rss_bytes) API.  I don't love this type of API because with the current separation of `procfs-core` and `procfs`, the interface ends up a little awkward and it would probably be easier for the API consumer to just get the page size and do the multiplication themselves.  But it's there.

procfs-core/src/types.rs
```diff
@@ -0,0 +1,79 @@
...
+
+        impl std::fmt::Display for $newtype {
+            fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+            std::fmt::Display::fmt(&self.0, f)
+            }
+        }
+    }
+}
+
+wrap_numeric!(Pages, u64, "A quantity of pages of memory");
```
eminence, line 62: I think this is missing some bits for handling the serde stuff:


```
error[E0277]: the trait bound `types::Pages: Deserialize<'_>` is not satisfied                                                                                                                
    --> procfs-core/src/buddyinfo.rs:20:18                                                     
     |                                                                                                                                                                                        
20   |     free_chunks: HashMap<Pages, u64>,                                                   
     |                  ^^^^^^^^^^^^^^^^^^^ the trait `Deserialize<'_>` is not implemented for `types::Pages`

```

PR #350, opened 2025-06-02
Add /proc/pid/syscall parsing support

jebradbury39:
Manpage for this file:

```
/proc/[pid]/syscall (since Linux 2.6.27)
              This file exposes the system call number and argument registers for the system call currently being executed by the process, followed by the values of the stack pointer and program counter registers.  The values of all six argument registers are exposed, although most system calls use fewer registers.

              If the process is blocked, but not in a system call, then the file displays -1 in place of the system call number, followed by just the values of the stack pointer and program counter.  If process is not blocked, then the file contains just the string "running". 
```

This file exists for tasks/threads as well.

### DIFF
--- a/procfs-core/src/process/syscall.rs
+++ b/procfs-core/src/process/syscall.rs
@@ -19,7 +19,6 @@ pub enum Syscall {
         /// Note that the rest of the values are still filled in.
         syscall_number: i64,
         /// The argument registers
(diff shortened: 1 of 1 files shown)


eminence:
Sorry for taking so long to notice and review this.  This looks good to me, thanks for the contribution

PR #359, opened 2025-08-31
Update missing protocol support in /proc/net

propertone:
These are needed for lsof feature parity support in [procfd](https://github.com/deshaw/procfd) (rust lsof replacement)


eminence:
Great, thank you for this update

PR #360, opened 2025-09-07
fix broken links to examples in documentation

Abstract-Everything:


Abstract-Everything:
Hi, I noticed some broken links from the documentation so I just pushed this to fix them, faster than creating an issue with a description :).

I did not check whether Rust has some kind of functionality which verifies that the links are not broken so that this does not happen again, although I doubt that it exists as it would only check it at compile time and the site can always go down later, although it would at least cover user mistakes I guess.

eminence:
Thanks for the fixes!

PR #363, opened 2025-12-30
feat(sys/kernel): add hostname and domainname

tastypackets:
Add `sys/kernel/hostname` and `sys/kernel/domainname` support.

Added a new line check on hostname & domainname which are used as a terminator like sysctl. Without the new line terminator trying to clear the values does not work. For example writing `""` to domainname is ignored, but writing `"\n"` does clear the domainname.

Related docs: https://man7.org/linux/man-pages/man5/proc_sys_kernel.5.html


procfs/src/sys/kernel/mod.rs
```diff
@@ -289,6 +290,48 @@ impl FromStr for BuildInfo {
...
+    };
+
+    write_value("/proc/sys/kernel/domainname", value)
+}
+
+/// Returns hostname
+///
+/// See also [set_hostname](crate::sys::kernel::set_hostname)
+///
+/// This is taken from `/proc/sys/kernel/set_hostname`
```
tastypackets:
```suggestion
/// This is taken from `/proc/sys/kernel/hostname`
```

tastypackets:
@eminence Hey would it be possible to get a review on this PR? I was needing hostname and domain for a project I'm working on.

eminence:
Sure thing, I'll try to take a look this weekend

eminence (changes requested):
In general looks fine, but two small comments:

* Do we really need these two examples?  They're not really doing much.
* Can you add a `cfg` check in the new tests so that they won't set a hostname unless some feature flag is enabled?  In general I don't want changes to be made to the running system (and these tests probably require root to run successfully?)

procfs/src/sys/kernel/mod.rs
```diff
@@ -570,6 +613,28 @@ mod tests {
         let _ = BuildInfo::current().unwrap();
     }
 
+    #[test]
+    fn test_set_and_get_domainname() {
+        let current_domain = domainname().unwrap();
+        let test_domain = String::from("test-host");
+
+        let test_domain = "test.local".to_string();
```
eminence: One of these variables is not needed

tastypackets: Nice catch think I copy pasted something and wasn't paying attention. 😓

PR #368, opened 2026-04-27
fix: correctly decode WSL2 Docker mounts

yuvald-sweet-security:
Hello,
My WSL2 installation with Docker Desktop has the following mount line, which causes this crate to panic:
`C:\134Program\040Files\134Docker\134Docker\134resources /Docker/host 9p rw,noatime,aname=drvfs;path=C:\Program Files\Docker\Docker\resources;symlinkroot=/mnt/,cache=5,access=client,msize=65536,trans=fd,rfd=3,wfd=3 0 0`

(note that the spec field is correctly mangled, but the `path` mntop is not and has spaces, causing the split by space strategy to fail)

Is this fix good? Do you prefer to do something else?

### DIFF
--- a/procfs-core/src/mounts.rs
+++ b/procfs-core/src/mounts.rs
@@ -31,12 +31,26 @@ impl super::FromBufRead for Vec<MountEntry> {
 
         for line in r.lines() {
             let line = expect!(line);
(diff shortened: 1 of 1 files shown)


eminence:
LGTM, thank you!

PR #369, opened 2026-05-02
fix possible panic in test mode

zhao-gang:
It's possible some values in /proc/meminfo would be abnormally big which would overflow the multiply calculation.

Didn't find related link on Internet but my Ubuntu 24.04 once showed incorrect SwapFree size which caused panic in procfs.

```
MemTotal:       32405976 kB
MemFree:         2062608 kB
MemAvailable:    5264828 kB
Buffers:           78156 kB
Cached:          6070016 kB
SwapCached:            0 kB
Active:          8976496 kB
Inactive:       17864952 kB
Active(anon):    8207228 kB
Inactive(anon): 15150596 kB
Active(file):     769268 kB
Inactive(file):  2714356 kB
Unevictable:     1696048 kB
Mlocked:             112 kB
SwapTotal:      67108860 kB
SwapFree:       18446744073708600176 kB
...
```

### DIFF
--- a/procfs-core/src/meminfo.rs
+++ b/procfs-core/src/meminfo.rs
@@ -3,12 +3,12 @@ use super::{expect, from_str, ProcResult};
 use serde::{Deserialize, Serialize};
 use std::{collections::HashMap, io};
 
-fn convert_to_kibibytes(num: u64, unit: &str) -> ProcResult<u64> {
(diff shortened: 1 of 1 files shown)


eminence:
> SwapFree:       18446744073708600176 kB

18000 exabytes sure is a lot of swap :)


This change looks good to me, thank you

PR #371, opened 2026-06-11
Fix a bunch of spelling mistakes

weriomat:
This is more or less a drive by fix. 
I was reading the documentation and noticed one mistake. Afterward I turned to the [typos](https://github.com/crate-ci/typos) spell checker for finding the other mistakes ;)

The change in the bit flag might be breaking if anyone is using this directly. I can also drop the commit if necessary.

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -51,7 +51,7 @@ Under the following license:
 
 Copyright 2003 Binh Nguyen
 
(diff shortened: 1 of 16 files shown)


eminence:
This is great, thank you.  Let's drop the bitflags typo fix for now, but if you want to open a dedicated PR with just that one change, I'll merge it later when it looks like there might be another reason for a semver bump

weriomat:
Dropped the commit, PR is coming :)

